PR #56, opened 2010-12-28
EscapeMode.none and starttag detection 

michael-simons:
Hi Jonathan,
i've implemented an EscapeMode.none that disables the escaping of text nodes. Yeah, i know it can be dangerous but there are cases, when it's needed.

Second thing i changed is the detection of tags. To my knowledge there is no valid tag that starts with a letter. With the change the parser creates better documents.

It would be very kind of you if you review my changes and pull.

Kind regards,
Michael.


zazi:
I'm also interested in EscapeMode.none. Any chance to get this merged into a more recent state of jsoup?

jhy:
I'm never going to implement EscapeMode.none as it only leads to broken parse trees.

PR #80, opened 2011-03-18
Added support for XPath 1.0

btd:
From test you can see usage.


ik-j:
I can get cssSelector using jSoup but am not able to get xpath:idrelative, xpath:attributes, xpath:location.
Though I am able to get xpath fixed with the code below:

StringBuilder absPath=new StringBuilder();
                    Elements parents = e.parents();

                    for (int j = parents.size()-1; j >= 0; j--) {
                        Element element = parents.get(j);
                        absPath.append("/");
                        absPath.append(element.tagName());
                        absPath.append("[");
                        absPath.append(element.siblingIndex());
                        absPath.append("]");
                    }

Help would be appreciated :)

PR #146, opened 2011-12-01
Cookie parsing

atoxic:
This change makes HttpConnection.Response try to parse the cookie with [HttpCookie](http://docs.oracle.com/javase/6/docs/api/java/net/HttpCookie.html) first and adds the cookie if it hasn't expired.  If HttpCookie can't parse it, then it falls back to the old method of parsing the cookie (including ignoring "expires" and other stuff).

I suppose that a full solution might require having the fallback method parsing the cookie into a map or something, but this is a start for spec-compliant cookies.


jhy:
Thanks, implemented similar with #1476

PR #173, opened 2012-03-12
Add an additional method to retreive an Elemnet as String

agaengel:
Hi,

to get an String Representation of an Element I added functionality to preserve blockElements and br as linebreaks in the returning String.

Greets,
Alexander


jhy:
Thanks, and sorry for the late review. Similar functionality was implemented in the `Element#getWholeText()` method.

PR #226, opened 2012-07-26
Set your own default charset

ile:
UTF-8 charset was not quite enough for me in all cases, so I made a version where you can set your own default charset.


jhy:

Issue #296, opened 2013-02-11
TextNode.isBlank() ignores Unicode whitespace 


mingfai:
TextNode.isBlank() treats only HTML whitespace as blank, instead of all Character.isWhitespace() characters as blank. So some unicode whitespace, such as '\u3000' in a text node will result as !isBlank(). This contradict with typical Java language usage, IMHO.

TextNode.isBlank is being used by other methods such as hasText(), so the change, if accepted, should be applied in the original isBlank.

jhy:
(Closing out old, dormant bugs. If you are still impacted by this, please reopen & vote.)

PR #301, opened 2013-02-20
https://github.com/jhy/jsoup/issues/296

mingfai:
Notice that I didn't change the definition of the original StringUtils.isWhiteSpace() as that was used for HTML parsing and other purpose. The change is applied to StringUtils.isBlank() only.

I made an additional change to rename the original isWhiteSpace to isHTMLWhiteSpace. I propose to make such change as there are two different usages of "whitespace" in the project.


mingfai:
This pull request is too old and irrelevant. I'm closing it.

Issue #249, opened 2012-10-07
JSoup 1.6.0+ (incl. 1.7.1) : <form> tag parsing failure when form is within a <table>


dma:
Hi, I seem to have encountered a serious bug in JSoup 1.6.0 - 1.7.1. Forms are not parsed correctly if they're in tables. JSoup misses nodes. This can be reproduced with the following simple HTML:

```html
<html>
<body>  
  <table>
      <form action="/hello.php" method="post">
      <tr><td>User:</td><td> <input type="text" name="user" /></td></tr>
      <tr><td>Password:</td><td> <input type="password" name="pass" /></td></tr>
      <tr><td><input type="submit" value="login" /></td></tr>
   </form>
  </table>
</body>
</html>
```

Let's try querying for all forms:

```java
package com.subgraph.JSoupTest;

import java.io.IOException;

import org.jsoup.Jsoup;
import org.jsoup.nodes.Document;
import org.jsoup.select.Elements;

public class JSoupTest {
        public static void main(final String[] args) throws IOException {
                Document doc = Jsoup.connect("http://www.subgraph.com/rtest/fds2.html").userAgent("Hello").timeout(5000).get();
                Elements forms = doc.select("form");
                System.out.println(forms);
        }

}
```

Here is the output, except the elements inside the form are missing:

```
<form action="/users/login.php" method="POST"></form>
```

When similar HTML without the table is parsed, such as this:

```html
<html>
  <body> 
      <form action="/hello.php" method="post">
      User:<input type="text" name="user" />
      Password:<input type="password" name="pass" />
      <input type="submit" value="login" />
   </form>
  </body>
</html>  
```

The output is as expected:

```
<form action="/users/login.php" method="POST">
  Username : 
 <input type="text" name="username" /> Password :
 <input type="password" name="password" /> 
 <input type="submit" value="login" />
 <a href="/users/register.php">Register</a> 
</form>
```

I've tested this with 1.5.2, and 1.5.2 produces the expected output when the form is in a table, so the bug must have somehow been introduced in 1.6.0.

jhy:
This was fixed in 1.7.3+ with FormElements. See https://github.com/jhy/jsoup/commit/834d3142304e0927dc659791da6a4521f271d7fd

With that change it works like @PiWizard pointed out Chrome / other browsers do.

```
FormElement form = (FormElement) doc.select("form").first();
List<Connection.KeyVal> data = form.formData();
```

PR #302, opened 2013-02-23
Address #249 + some additional bug fixes - see commits

dma:
Hi jhy,

I made some changes to improve the FormElements implementation. Missing elements get included, and there was one possibility of duplication that I fixed.


jhy:
Thanks, and sorry for the late reply. This is similar to some other work with FormElements

PR #306, opened 2013-03-17
Document.sourceUri()

boomkap:
Add an accessor method sourceUri to retrieve the final (in case of redirects) URI from which the document was parsed.


jhy:
(Not sure why this is not marked as merged. It is.)

boomkap:
Cool.

PR #324, opened 2013-04-29
Add hasOwn selector syntax; Allow NodeVisitors to remove the node they're visiting

jloomis:
Add hasOwn selector syntax:
Mirroring matchesOwn & containsOwn, allow selection of an element which directly contains the given selector.

Allow NodeVisitors to remove the node they're visiting:
Useful for example when pruning elements - rather than building up a list of elements to remove then removing them you can remove them during traversal.


jhy:

PR #340, opened 2013-06-26
HttpConnection - Redirect with '#' char on Android

stephdz:
I've noticed that, for some weird reason, on Android. when a redirect response contains a '#' (for exemple : http://www.20minutes.fr/article/1180291/1180291#xtor=RSS-145), HttpURLConnection fails and returns a 404 error.

So, I've fixed HttpConnection in order to extract only the left part of the URL, and then, it works.


jhy:

PR #343, opened 2013-07-05
Provide ability to disable SSL certificates check.

aakhmerov:
There are sometimes cases in business appliance where one really wants to disable SSL checks, for instance during integration with third party systems it is not always possible to force them to update certificates used for https. 

Unit test is provided under integration tests section using washington university host. I would suggest providing proper endpoint at jsoup.org for that and refactoring unit test. 

11.11
- refactored code to do checks before aggressifve cast to HTTPSUrlConnection
- added proper unit test


zeynepsu:
Was this change abandoned? The functionality to disable SSL checks is not present in the current version of Jsoup 1.8.3 which was published on August 2015 in maven.

PR #384, opened 2014-01-09
Add Prototype Style DOM Navigation

epochcoder:
the `Element` class now has the ability to quickly go `up`/`down`/`next`/`previous` according to the specified CSS selector

Example:

``` java
Document doc = Jsoup.connect("https://en.wikipedia.org/wiki/Java_(programming_language)").get();

Element fullTableOfContentsWhileNavigating = doc
            .down("li.toclevel-1.tocsection-13")
            .previous("li")
            .down("a[href=#A_more_comprehensive_example]")
            .up("div#toc");
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
   		<name>The MIT License</name>
   		<url>http://jsoup.org/license</url>
   		<distribution>repo</distribution>
-  	</license>  
+  	</license>
   </licenses>
   <scm>
   	<url>http://github.com/jhy/jsoup</url>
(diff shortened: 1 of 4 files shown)


epochcoder:
Hi @jhy, Is there any more info you might require?

jhy:
(Closing old PRs; thanks for your interest and apologies for the late review. I don't plan to merge this; mostly because the names aren't familiar to me or I think most others, not being users of Prototype. We do have `closest` which I think aligns to `up`? And `select()` which is `down()`?. If the siblings selectors are useful we could add those too.)

PR #435, opened 2014-07-08
do not send expired cookies

itay1336:
Some websites return multiple cookie values using the same cookie name.
It seems like browsers do not forward the expired cookies in the next request when a redirection occurs.
I started seeing this problem when I was trying to parse some article on the NYT website.
One of the responses had a cookie called "NYT-S" and it had multiple values, one the values had an expire time (1970) and another value (2015). When Jsoup built the request it was overriding the non expired cookie value with the expired cookie value which cause the NYT site to not be parsable.
I also changed cookie parsing logic to use HttpCokie class since there are different versions of HTTP cookies and I didn't want to write the same code which was already implemented in Java.
Please see:
http://docs.oracle.com/javase/6/docs/api/java/net/HttpCookie.html

### DIFF
--- a/src/main/java/org/jsoup/helper/HttpConnection.java
+++ b/src/main/java/org/jsoup/helper/HttpConnection.java
@@ -1,21 +1,34 @@
 package org.jsoup.helper;
 
-import org.jsoup.Connection;
-import org.jsoup.HttpStatusException;
-import org.jsoup.UnsupportedMimeTypeException;
-import org.jsoup.nodes.Document;
-import org.jsoup.parser.Parser;
-import org.jsoup.parser.TokenQueue;
-
-import java.io.*;
+import java.io.BufferedInputStream;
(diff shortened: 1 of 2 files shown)


pboos:
Nice pull request. I would like it to be merged in.

But one problem:
Currently it is built for java 1.5. So "mvn package" actually fails because of the checks for making sure it is java 1.5 compatible (animal-sniffer-maven-plugin -> java15). As well as target we have 1.5 in the pom.xml. So depending on if android should be supported or not, this should be changed or HttpCookie should not be used. Either way :).

But this problem you are fixing needs to be fixed!! And either this one needs to be merged or another version of this fix.

itay1336:
I will try to build it and fix the issues. thanks!

jhy:
Sorry, can't take this in as HttpCookie only exists in 1.6 up. Need to support 1.5

PR #450, opened 2014-08-20
added proxy

larryhu:
add a proxy setting to the request.


jhy:
Thanks - implemented a similar setting

PR #457, opened 2014-09-02
add charset(String charset)  in Request

gwuhaolin:
add charset(String charset) method in Request,so can set encode charset to encode request data


jhy:
Thanks - similar implementation got merged a while back

PR #465, opened 2014-09-22
Allow overriding processResponseHeaders of HttpConnection

monzou:
Hi.

I want to customize response headers of HttpConnection. So I've changed the `HttpConnection#Request` and `HttpConnection#Response` to not `static`, and changed some methods to `protected` for overriding.

If there aren't any problems, please merge this PR.


jhy:
Thanks, and sorry for the late reply. I have been reluctant to make the implementation more open as it has seen a lot of changes.

One approach that might be useful would be to add a callback / event type system for different stages of the request / response flow.

Issue #467, opened 2014-10-08
Allow for node serialization via OutputStream


schmid-michael:
As discussed  <a href="https://groups.google.com/forum/#!topic/jsoup/f2cOnH43dqE" target="blank">here</a> it should be possible to serialize a Document/Node via an OutputStream rather than its String representation.

jhy:
Merged with the recent changes - https://github.com/jhy/jsoup/pull/470/

PR #470, opened 2014-10-11
Changes for issue #467

jhy:


caseyoneill:
Was there any reason that this wasn't accepted? It seems like serialization of the Node objects is quite important.

jhy:
Thanks! Merged.

Issue #466, opened 2014-10-06
viewBox converted into viewbox in svg files


Sharcoux:
Hi,
Using JSoup, the attribute "viewBox" is converted into "viewbox" and this causes SVGPanel from SVGSalamander not recognizing it. I don't know if I should consider this a problem with JSoup or SVGSalamander, but I think that Jsoup should leave quiet the viewBox attribute if possible.

Thanks anyway for your useful library !

jhy:
You can use ParseSettings to control case normalization now. Thanks.

Issue #407, opened 2014-03-28
Tag to lower case


lgabardos:
Hi,

I planned to use JSoup in order to clean my XML.

But, with this library, it always transform tags in lowercase... and I really need to keep my tags like they were.

So, I suggest a  thing:
In org.jsoup.parser.Tag, add 

public static boolean TAG_TO_LOWERCASE = true;

And after, when the tagName is lowercased, we have to check if TAG_TO_LOWERCASE is true, like that

if (Tag.TAG_TO_LOWERCASE) {
	tagName = tagName.toLowerCase();
}

If the user want to parse (or clean) XML, before processes, we just need to set 
Tag.TAG_TO_LOWERCASE = false;

And, it's good.

What do you think about that ?

Elinow:
Agree it is required. By specification XML tags are case sensitive.

jhy:
You can now control case sensitivity / normalization with ParseSettings. https://github.com/jhy/jsoup/commit/e38b916ba1ebd3f5c00a6b6b8e77a4cdfea2800f

Issue #344, opened 2013-07-06
jsoup's XML parser lowercases tags


Raynes:
A user of a library of mine that uses jsoup under the hood recently submitted an issue pointing out that XML tags are being lowercased. I did a little experimentation and it appears that JSoup is doing the lower casing.

https://github.com/Raynes/laser/issues/25

Here is an example

```clojure
user> (Jsoup/parse "<Test xmlns:Ex=\"yo\"/>" "" (Parser/xmlParser))
#<Document <test xmlns:ex="yo" />>
```

jhy:
Support for case preservation is in https://github.com/jhy/jsoup/commit/e38b916ba1ebd3f5c00a6b6b8e77a4cdfea2800f

Issue #328, opened 2013-05-06
node.attributes() converts node keyName to all lowercase.


nitinkcv:
Suppose i have the following xml node:
`<foo bar="test" fooBar="test1" barFoo="test2">bar</foo>`

```
Iterator<Attribute> itr = node.attributes().iterator();
while(itr.hasNext()){
    
    Attribute attr = itr.next();
    String attrKey = attr.getKey(); // gives foobar, barfoo instead of fooBar,barFoo
}
```

jhy:
You can use ParseSetttings to control case normalization now.

surinaidu:
May I know how to use ParseSettings before parsing the xml data?

krystiangorecki:
If you use Parser.xmlParser() then the case is preserved by default. Setting it manually can be done this way:
```import java.util.Iterator;

import org.jsoup.Jsoup;
import org.jsoup.nodes.Attribute;
import org.jsoup.nodes.Document;
import org.jsoup.nodes.Element;
import org.jsoup.parser.ParseSettings;
import org.jsoup.parser.Parser;

public class JsoupIssue328 {

    public static void main(String[] args) {
        Parser xmlParser = Parser.xmlParser();
        // next line is redundant because ParseSettings.preserveCase is default for xmlParser!
        xmlParser.settings(ParseSettings.preserveCase);
        Document doc = Jsoup.parse("<foo bar=\"test\" fooBar=\"test1\" barFoo=\"test2\">bar</foo>", "", xmlParser);
        Element node = doc.select("foo").first();
        Iterator<Attribute> itr = node.attributes().iterator();
        while (itr.hasNext()) {
            Attribute attr = itr.next();
            String attrKey = attr.getKey();
            System.out.println(attrKey);
        }
    }

}
```

PR #480, opened 2014-10-21
Allow case-sensitive retrieval of element and attribute names.

tipabu:
Addresses #29, #272, #328, #344, #407, and #466.

`Element`s and `Attribute`s each have new getters (`tagName` and `getKey`, respectively) that take a boolean (`preserveCase`) as an argument. If true, the original string used to construct the object will be returned; otherwise, the lower-cased version will be returned, as before. Additionally, a new `preserveCase` option has been added to `OutputSettings`: this is referenced when rendering content using `outerHtml()` and the like. When rendering, the closing tags for elements will always match the casing of the opening tag.

There are two implementation details worth noting, however: First, `Tag`s, `Element`s, and `Attribute`s now retain both the original tag/attribute name and a lower-case version of it. This has some benefits at the cost of an increased memory footprint:

 1. Existing code that makes direct reference to `tagName` and `key`
    didn't need to change.
 2. Case-insensitive comparison (for end-tag matching and element
    selection, for example) is somewhat faster than repeatedly calling
    `.equalsIgnoreCase()` on the same string.

Second, HTML tags that are not already all-lower-case will be treated like unrecognized XML tags, with new `Tag` objects created for each. Like other unknown `Tag`s, these will `.equals()` but not `==` other `Tag`s with the same (case insensitive) name. Their other properties (`canContainBlock`, `canContainInline`, etc.) will be copied from the registered version. This has two downsides:

 1. The memory footprint will again increase, though it should be no
    worse than parsing a comparably-sized XML document.
 2. Existing code that relies on strict equality of HTML tags with the
    same name will break. This is unavoidable if we wish to preserve,
    for example, both `<p>` and `<P>` in a document, though we may still
    want to make `Tag.valueOf("P") == Tag.valueOf("P")`. It should be
    noted, however, that `Element` doesn't offer a comparable guarantee.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Attribute.java
+++ b/src/main/java/org/jsoup/nodes/Attribute.java
@@ -18,10 +18,12 @@ public class Attribute implements Map.Entry<String, String>, Cloneable  {
     };
 
     private String key;
+    private String keyOrig;
     private String value;
 
     /**
      * Create a new attribute from unencoded (raw) key and value.
+     *
      * @param key attribute key
      * @param value attribute value
      * @see #createFromEncoded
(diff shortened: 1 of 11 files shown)


cosmin:
+1 for this request. I have an application I need to deal with that for some reason managed to breaks when the case changes on the output, even though it is supposed to output HTML which ought to be case insensitive.

The merge conflicts here are minimal. Only an issue with test files, and even there the problem is that git became confused about the placement of a few `}` to properly end one test and start another. I'm happy to resubmit the pull request with this cleaned up if that would make this easier.

cuantico:
+1

zazi:
+1 (this enhancement looks also a bit more comprehensive then #527 ;) )

manolo:
+1

jhy:
Thanks for this. I have implemented ParseSettings which allows case normalization options for Tags and Attributes - similar concept but slightly different implementation.

PR #486, opened 2014-10-27
Update meta charset

offa:
This pull request attends the todo-comment according the update of documents meta charset tag.

If the charset is changed, the charset meta-tag of the document is slected and updated (if existing).

An improvement could be an overload `charset()` method, where the caller can choose if the meta tag will be updated too.

A test is included.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Document.java
+++ b/src/main/java/org/jsoup/nodes/Document.java
@@ -18,6 +18,7 @@ public class Document extends Element {
     private OutputSettings outputSettings = new OutputSettings();
     private QuirksMode quirksMode = QuirksMode.noQuirks;
     private String location;
+    private boolean updateMetaCharset = false;
 
     /**
      Create a new, empty Document.
@@ -187,12 +188,6 @@ public class Document extends Element {
     }
 
     @Override
-    public String html() {
-        ensureMetaCharset();
-        return super.html();
-    }
-
-    @Override
     public String outerHtml() {
         return super.html(); // no outer wrapper tag
     }
@@ -212,6 +207,23 @@ public class Document extends Element {
     public String nodeName() {
         return "#document";
     }
+    
+    public void charset(Charset charset) {
+        outputSettings.charset(charset);
+        ensureMetaCharset();
+    }
+    
+    public Charset charset() {
+        return outputSettings.charset();
+    }
+    
+    public void updateMetaCharset(boolean update) {
+        this.updateMetaCharset = true;
+    }
+    
+    public boolean updateMetaCharset() {
+        return updateMetaCharset;
+    }
 
     @Override
     public Document clone() {
@@ -221,7 +233,7 @@ public class Document extends Element {
     }
     
     private void ensureMetaCharset() {
-        if( outputSettings.updateMetaCharset() ) {
+        if( updateMetaCharset == true ) {
             Element metaCharset = select("meta[charset]").first();
             
             if( metaCharset != null ) {
@@ -255,7 +267,6 @@ public class Document extends Element {
         private boolean outline = false;
         private int indentAmount = 1;
         private Syntax syntax = Syntax.html;
-        private boolean updateMetaCharset = false;
 
         public OutputSettings() {}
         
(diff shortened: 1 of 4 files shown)


jhy:
@offa sorry for the super late review on this. Are you able to re-base this against head? The diff has a load of later changes from other commits.

offa:
Merge is done. While we're at it i have over-worked the tests for this.

offa:
@jhy i have merged the latest changes from today as well..

src/main/java/org/jsoup/nodes/Document.java
```diff
@@ -206,6 +209,79 @@ public Element text(String text) {
...
+     * <li><b>Html:</b> <i>&lt;meta charset="CHARSET"&gt;</i></li>
+     * <li><b>Xml:</b> <i>&lt;?xml version="1.0" encoding="CHARSET"&gt;</i></li>
+     * </ul>
+     * 
+     * @param charset Charset
+     * 
+     * @see #updateMetaCharset(boolean) 
+     * @see OutputSettings#charset(java.nio.charset.Charset) 
+     */
+    public void charset(Charset charset) {
```
jhy: I think that it would be better to turn on updateMetaCharset when this method is hit. If the user didn't want it updated, they wouldn't hit this method. They would just use the Output settings.

src/main/java/org/jsoup/nodes/Document.java
```diff
@@ -206,6 +209,79 @@ public Element text(String text) {
...
+     * 
+     * <p>If set to <tt>false</tt> <i>(default)</i> there are no elements
+     * modified.</p>
+     * 
+     * @param update If <tt>true</tt> the element updated on charset
+     * changes, <tt>false</tt> if not
+     * 
+     * @see #charset(java.nio.charset.Charset) 
+     */
+    public void updateMetaCharset(boolean update) {
```
jhy: Maybe call this updateMetaCharsetElement ?

src/main/java/org/jsoup/nodes/Document.java
```diff
@@ -213,6 +289,84 @@ public Document clone() {
...
+     * 
+     * <p><b>Elements used:</b></p>
+     * 
+     * <ul>
+     * <li><b>Html:</b> <i>&lt;meta charset="CHARSET"&gt;</i></li>
+     * <li><b>Xml:</b> <i>&lt;?xml version="1.0" encoding="CHARSET"&gt;</i></li>
+     * </ul>
+     */
+    private void ensureMetaCharset() {
+        if( updateMetaCharset == true ) {
```
jhy: Can you check syntax (space) formatting on this? Expecting if (x == y) {...

src/main/java/org/jsoup/nodes/Document.java
```diff
@@ -274,7 +428,6 @@ public Charset charset() {
          * @return the document's output settings, for chaining
          */
         public OutputSettings charset(Charset charset) {
```
jhy: Maybe I'm missing it; how does changing this update the charset element? I.e. if I change the output settings after the parse (and normalize) is completed.

src/test/java/org/jsoup/nodes/DocumentTest.java
```diff
@@ -157,4 +162,229 @@
...
+        final String htmlCharsetUTF8 = "<html>\n" +
+                                        " <head>\n" +
+                                        "  <meta charset=\"" + charsetUtf8 + "\">\n" +
+                                        " </head>\n" +
+                                        " <body></body>\n" +
+                                        "</html>";
+        assertEquals(htmlCharsetUTF8, doc.toString());
+        
+        Element selectedElement = doc.select("meta[charset]").first();
+        assertEquals(charsetUtf8, doc.charset().displayName());
```
jhy: I believe every instance of displayName() should be name(). Latter is canonical, but former will / can vary depending on Locale. http://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html#displayName()

jhy:
@offa this is looking great! Looking forward to bringing it in. I added some notes, can you please review?

jhy:
Let me have a think about the content-type meta field. I remember we kind of handle that in DataUtils. Could be that we pull out a help method for deciding which to use.

offa:
> @offa this is looking great! Looking forward to bringing it in. I added some notes, can you please review?

@jhy thanks for the comments.

jhy:
Cool bananas. Is that ready to go then? 

What about my Q on DocumentOutputSettings?

offa:
> Maybe I'm missing it; how does changing this update the charset element? I.e. if I change the output settings after the parse (and normalize) is completed.

This does will not update the elements. This was implemented earlier but then removed (as discussed somewhere above).

> Is that ready to go then? 

If there are no further issues and travis gives his Ok i think so.

jhy:
Cool! Merged. Thanks for your work on this, much appreciated.

offa:
:+1:

PR #487, opened 2014-10-28
Refactor the Elements collection to extend instead of delegate

cromoteca:
Maybe it's just a personal preference, but if this collection is used a lot, maybe skipping the delegation can improve performance.


jhy:
Thanks!

jhy:
(BTW I pulled it because it reduces code and simplifies. Not sure about any perf benefit.)

Issue #155, opened 2011-12-11
Can't specify POST request encoding


archislav:
For example, I want to make POST request for http://mysite.com with parameter 'a' equals to bookName variable. 

Document searchPage = Jsoup.connect("http://mysite.com").data("a", bookName).post();

That's OK for latin letters, but it fails, when I try to use it with cyrilic letters in windows-1251, bookName is URLencoded with utf-8 encoding in POST data.

You can look at HttpConnection.writePost method, it uses DataUtil.defaultCharset = "UTF-8".

So it will be great if you provide an ability to specify encoding for request. 
For example:

Document searchPage = Jsoup.connect("http://mysite.com").data("a", bookName).requestEncoding("windows-1251").post();

jhy:
Fixed with that merge, thanks.

PR #491, opened 2014-11-04
Update to allow for specification of form post data charset.

andyphillips404:
Updates to Connection and Request interfaces and HttpConnection implementation.   This allows for the specification of the charset used for application/x-www-form-urlencoded post data character set, which is needed for more robust support.   The implementation was done by adding a postDataCharset method to the request object with the default value still maintaining the DataUtils.defaultCharset.   This implementation should have minimum effect on current applications as the default is still UTF-8 but allows the setting of ISO-8859-1 which appears frequently in use on the web.  I've also added the charset= parameter to the header setting to allow the server to be aware of the charset used, although I've found this isn't always honored...

Fixes issue #155

### DIFF
--- a/src/main/java/org/jsoup/Connection.java
+++ b/src/main/java/org/jsoup/Connection.java
@@ -211,6 +211,13 @@ public interface Connection {
     public Connection parser(Parser parser);
 
     /**
+     * Sets the default post data character set for x-www-form-urlencoded post data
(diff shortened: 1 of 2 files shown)


jhy:
Thanks, looks great

andyphillips404:
Your welcome, love the product.

PR #493, opened 2014-11-13
Delete redundant nullcheck code in HttpConnection.java

SubiRUN:
String cookieVal = cd.consumeTo(";").trim();
 // cookieVal will never be null
 if (cookieVal == null) // should be deleted


jhy:
Thanks. This was picked up by another cleanup.

PR #496, opened 2014-11-24
Double quotes in text nodes does not get escaped when using XHTML escaping mode

lchngr:
I think this is a simple typo. Double quotes are not escaped when using XHTML mode. &<> characters are fine.


jhy:
@lchngr quotes in character data don't need to be escaped. In xml, html, or xhtml. E.g. see http://xml.silmaril.ie/specials.html

PR #517, opened 2015-01-03
Fixed JavaDoc errors and warnings for JDK8 with stricter doc rules

offa:
Building Jsoup with JDK8 causes a long list of JavaDoc *warnings (5)* and *errors (86)*. Maven reports a failed build for Javadoc, however it's still working and correct. The reason is, that JDK8 introduced stricter rules for html formating.

This pull request fixes *all* of these errors / warnings so a clean and successful build is possible with JDK8. The documentation's formating itself is kept.


jhy:
Thanks! Did you do that by hand (crikey) or did you find a tool?

offa:
By hand, but with the help of search & replace.

Issue #272, opened 2012-12-07
XML Attribute Names are converted to lower case, where they should stay as in the original input


edgar-philipp:
Hello,

in my current project I stumbled on this curious behaviour:

We use the following version of JSoup:

`<dependency>
	<groupId>org.jsoup</groupId>
	<artifactId>jsoup</artifactId>
	<version>1.7.2</version>
</dependency>`

```java
@Test
public void testJsoupAttributeNameCamelCase() {

        String xml = "<?xml version='1.0' encoding='UTF-8' standalone='no'?>" 
                              + "<rootNode attributeName='someValue' />";
        Document doc = Jsoup.parse(xml, "", Parser.xmlParser());
        Element root = doc.getElementsByTag("rootNode").get(0);
        Attribute first = root.attributes().asList().get(0);

        assertTrue("someValue".equals(first.getValue())); // pass
        assertTrue("attributeName".equals(first.getKey())); // fail
    }
```

Please have a look at this issue.

Thank you and kind regards
Edgar

JavaJeffG:
+1 to allow for case-sensitive retrieval.

I am reading in the contents of an HTML table which may or may not contain XML elements, e.g.,
     <td><firstName>Frank</firstName></td>

When I get the 'td' org.jsoup.nodes.Element, I cannot access the case sensitive contents:

  tdElement.text()        = Frank
  tdElement.html()       = <firstname>Frank</firstname>
  tdElement.toString() = <td><firstname>Frank</firstname></td>

This is part of a case-sensitive Selenium test where <firstName>Frank</firstName> needs to be on the web page.

How about tdElement.htmlRaw() that provides the raw text?

Thanks.

itonics-tbeauvais:
This is incredibly frustrating, is there a solution?

cppexpert:
Guys, any progress on this ?

roben:
This really is frustrating for an otherwise great library. Zazi seems to have provided a ready to use solution. Is there any reason why it has not been merged yet?

kamrankashef:
Please merge.  Case-sensitive nodeName() and getKey() would be extremely helpful.

jhy:
I have implemented a new feature ParseSettings, which allows the user to preserve tag and/or attribute name case at parse (and therefore output) time. For XML it defaults to preserving case, and lower case for HTML parsing.

https://github.com/jhy/jsoup/commit/e38b916ba1ebd3f5c00a6b6b8e77a4cdfea2800f

[...]

Issue #29, opened 2010-08-05
lower cased html attributes


aseychell:
As I already stated in a previous post, we are using JSTL tags (java custom tags) and we require the attributes to be camel cased to match some methods in our java code.  Is it possible to give an option to leave the attributes as they are and not modify them by making them lower case?

e.g. <abc:ourtag returnUrl="http://abc.com" /> does not change to <abc:ourtag returnurl="http://abc.com" />

Thanks!

jhy:
You can control case normalization with ParseSettings now - https://github.com/jhy/jsoup/commit/e38b916ba1ebd3f5c00a6b6b8e77a4cdfea2800f

PR #527, opened 2015-02-09
Add option to configure Jsoup's attribute normilization.

### LINKED ISSUE #328: node.attributes() converts node keyName to all lowercase.

dankurka:
Currently Jsoup turns all keys of attributes into lowercase.
This breaks Jsoup when being used with many modern web projects
such as Polymer (https://www.polymer-project.org/) since these
projects use camel case attributes in their code.

This patch adds an option to Jsoup so that users can disable
this kind of normaliztion.

fix #328, #272, #29

### DIFF
--- a/src/main/java/org/jsoup/JsoupOptions.java
+++ b/src/main/java/org/jsoup/JsoupOptions.java
@@ -1,0 +1,42 @@
+package org.jsoup;
+
+/**
+ The Jsoup options.
+
+ @author Daniel Kurka */
+public class JsoupOptions {
+
+  public static final JsoupOptions DEFAULT_OPTIONS =
+      new Builder().normalizeAttributes(true).build();
+
+  private boolean normalizeAttributes;
+
+  /**
+   Should Jsoup normalize attributes. This will turn all attribute keys into
+   lowercase.
+   */
+  public boolean shouldNormalizeAttributes() {
+    return normalizeAttributes;
+  }
+
+  /**
+   A builder to create a JsoupOptions object.
+  */
+  public static class Builder {
+    private JsoupOptions options = new JsoupOptions();
+
+    /**
+     Should we normalize attributes:
+
+     This would turn an uppercase attribute key into all lowercase.
+     */
+    public Builder normalizeAttributes(boolean normalize) {
+      options.normalizeAttributes = normalize;
+      return this;
+    }
+
+    public JsoupOptions build() {
+      return options;
+    }
+  }
+}
--- a/src/main/java/org/jsoup/Jsoup.java
+++ b/src/main/java/org/jsoup/Jsoup.java
@@ -1,11 +1,11 @@
 package org.jsoup;
 
+import org.jsoup.helper.DataUtil;
+import org.jsoup.helper.HttpConnection;
 import org.jsoup.nodes.Document;
 import org.jsoup.parser.Parser;
 import org.jsoup.safety.Cleaner;
 import org.jsoup.safety.Whitelist;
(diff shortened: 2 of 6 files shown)


slopetown:
+1

jDramaix:
+1

michaelvogt:
+1

dankurka:
friendly ping

alpapad:
+1

dankurka:
friendly ping

jeffgnpc:
+1

jhy:
@dankurka thanks for this, and sorry for the late reply.

My concern with this implementation is that the state of the option is made global to the whole VM. If users are running concurrently, and are setting this option depending on some other context (i.e. maybe per site in a scrape) they are going to hit race errors.

Did you consider implementing this in Document.OutputSettings? I think it could fit there reasonably well. While not purely only an output setting, we could document the behavior that changing that output setting also impacts the DOM queries and that should make sense to users. Or if not in Document.OutputSettings, just a switch in Document itself.

And we can supply a configuration to the Parser to enable it for the parse-time execution.

Also, I don't know that the setting is best named attribute normalization. There are other normalizations that still occur (e.g. space trimming). I'd be more literal / specific and call the option something like "lower case attributes".

src/main/java/org/jsoup/JsoupOptions.java
```diff
@@ -0,0 +1,42 @@
+package org.jsoup;
+
+/**
+ The Jsoup options.
+
+ @author Daniel Kurka */
+public class JsoupOptions {
```
sambathl: Hi Daniel,
              Please correct if I am wrong, why donot we use ENUM instead of class for JsoupOptions

Cheers!!!

dankurka:
@jhy I briefly looked at it in general tying this to the Document / Settings makes sense, but the Documen tinstance is not available in all places where there is currently a conversion happening (e.g. Attribute constructor).

Should I be looking into refactoring the call sites to make this happen?

dankurka:
Looked some more at this:

Since the parser already does lower case conversions (and we thus loose data), we can not really tie this to the Document / its settings, since you only get access to the Document after parsing.

A possible solution might be to remove the lower case conversions on write (construtors, parser, etc) and store the data as camel case. Then make the Output settings available to Attributes and convert if necessary on read. What do you think?

jhy:
I think we need to make it mostly a parse time change. I want to be careful
not to impact select or serialization time when implementing this feature.
Output settings, on second thoughts, is probably not the right place for it.

What's the scope of change if we make it a Parser option, and that
preserves case when building the Attributes. And set a flag in Document, so
Selectors know not to lowercase attribute queries? That might be the same
problem when pulling an Attribute by key though...

Let me take another read through and try to think of something. Ideas which
meet constraints of not a global VM switch, and don't impact perf, are very
welcome.

tipabu:
FWIW, I've got a similar change in #480, implemented by storing both the original and lower-case versions then using an Output setting to decide which to return. It should give an idea of the scope necessary to make it a Parser option, and may make a good starting point. Haven't had time to try to do that myself yet, though.

And yes, Attributes suffer from the same problem.

jhy:
I implemented a version of this with ParserSettings. Thanks.

PR #533, opened 2015-03-06
Use platform independent line endings in W3CDomTest.

torrlane:
The W3CDomTest simpleConversion test fails on windows since it is using "\n" as the
line ending. Using a platform independent line ending i.e. \r\n on windows
or \n on unix, fixes it.


jhy:
Thanks for the fix!

Issue #539, opened 2015-03-17
bump invalid head tags down into body?


jaredstehler:
Would it be possible to automatically move invalid head tags down into the body, without ending head processing as is the current behavior?

My situation is that I've got a large amount of created content which includes the ```<noscript><img></noscript>``` tracking pixel in the page header, as is recommended by Facebook (https://www.facebook.com/help/373979379354234). When I parse the page with jsoup prior to rendering, it closes the head at the point of the invalid tag and ends up bumping all of the other elements into the body (i.e. ```<link>``` tags).

Perhaps we could store invalid elements during the head parse, and dump them into the top of the body element?

jaredstehler:
Or as an alternative, convert the invalid tags into CDATA or equivalent?

Here's a simple test I did in chrome.

![screen shot 2015-03-17 at 9 59 53 am](https://cloud.githubusercontent.com/assets/2022218/6688792/80e643b6-cc8c-11e4-95ae-5bde209c569d.png)

PR #540, opened 2015-03-17
store invalid head/noscript tokens as character tokens

jaredstehler:
update the ```InHeadNoscript``` parser to insert invalid tokens as character nodes rather than prematurely ending the head node.

Fixes https://github.com/jhy/jsoup/issues/539


jhy:
Thanks!

PR #547, opened 2015-04-03
Addressing slow Element.outerHtml() on Android 4.4 and 5

gregko:
The big slowdown on Android 4.4 and 5 is the call to canEncode(c) in Entities.escape() method. Assuming that virtually any codepoint can be encoded in Unicode (e.g. UTF-8, UTF-16), added a check for encoder being Unicode and skipping this costly call. Speeds up Android 4.4 and 5 calls about 10 times.
Added some small other optimizations in a few other modules.

Greg

### DIFF
--- a/src/main/java/org/jsoup/nodes/Document.java
+++ b/src/main/java/org/jsoup/nodes/Document.java
@@ -191,6 +191,11 @@ public class Document extends Element {
         return super.html(); // no outer wrapper tag
     }
 
+    @Override
+    public String outerHtml(int initLength) {
+        return super.html(initLength);
+    }
+
     /**
      Set the text of the {@code body} of this document. Any existing nodes within the body will be cleared.
      @param text unencoded text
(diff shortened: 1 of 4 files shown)


jhy:
Thanks @gregko for your help in fixing this perf issue. I have committed a similar change, so won't pull this one.

Did you measure a performance improvement from intitializing the buffer size? I'm not sure how a user would know what size to specify. If it's a big improvement, maybe we could think of a heuristic to pre-calculate it? Seems that it should be conservative and err on under-estimating.

gregko:
Great, Jonathan, thank you for implementing this solution, will help me stay with your main code instead of working with my own fork!
I tried to measure if there is any performance improvement with initializing buffer size, and on an HTML file of about 200 kB length did not notice any improvement compared with the default buffer length (measured ms spent in doc.outerHtml() routine).

PR #551, opened 2015-04-13
Fix element ordering bug introduced by previous refactor

pfarrel:
Revert previous commit by jhy which introduced this reordering bug. Includes minimal repro used to bisect its introduction.


pfarrel:
Oops - meant to open against a different repo.

Issue #503, opened 2014-12-04
Allow collapsing of custom attributes


Artur-:
When using web components / custom attributes, the attributes will never be collapsed because there is a fixed set of collapsible attributes defined in Attribute.java. Because of this, output becomes unnecessary complex.

jhy:
Thanks, this is implemented with #555

PR #555, opened 2015-04-18
Add custom boolean attribute support #503

Legioth:
* Adds BooleanAttribute that writes out itself without a value
* Adds API in Element for setting boolean attributes
* Update parser to distinguish between no value and empty value

### DIFF
--- a/src/test/java/org/jsoup/nodes/ElementTest.java
+++ b/src/test/java/org/jsoup/nodes/ElementTest.java
@@ -294,6 +294,22 @@ public class ElementTest {
             assertEquals(i, ps.get(i).siblingIndex);
         }
     }
+    
+    @Test public void testAddBooleanAttribute() {
+        Element div = new Element(Tag.valueOf("div"), "");
+        
+        div.attr("true", true);
+        
+        div.attr("false", "value");
+        div.attr("false", false);
+        
+        assertTrue(div.hasAttr("true"));
+        assertEquals("", div.attr("true"));
+        
+        assertFalse(div.hasAttr("false"));
+ 
+        assertEquals("<div true></div>", div.outerHtml());
+    }    
 
     @Test public void testAppendRowToTable() {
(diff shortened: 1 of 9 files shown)


src/test/java/org/jsoup/nodes/ElementTest.java
```diff
@@ -294,6 +294,22 @@
...
+        div.attr("false", "value");
+        div.attr("false", false);
+        
+        assertTrue(div.hasAttr("true"));
+        assertEquals("", div.attr("true"));
+        
+        assertFalse(div.hasAttr("false"));
+ 
+        assertEquals("<div true></div>", div.outerHtml());
+    }
```
jhy: Can you assert their type here also?

src/test/java/org/jsoup/parser/AttributeParseTest.java
```diff
@@ -66,4 +66,16 @@
         Elements els = Jsoup.parse(html).select("a");
         assertEquals("&wr_id=123&mid-size=true&ok=&wr", els.first().attr("href"));
     }
+    
+    @Test public void parsesBooleanAttributes() {
```
jhy: Assert type here also

jhy:
Thanks, looks interesting. Can you check what the behavior should be for XML parsing and XML syntax output? Thinking that at least for XML syntax output, boolean attributes should actually output as empty string attributes.

Legioth:
From what I have found, the original XHTML specification only allowed the format `attribute="attribute"`.

Latest XHTML specs also allow `attribute=""` according to e.g. https://html.spec.whatwg.org/#boolean-attributes

For attributes defined as boolean using XML Schema, the format is `attribute="true"` or `attribute="1"`  (not compatible with XHTML version).

Legioth:
Carefully bumping this.

jhy:
Thanks! Merged.

PR #559, opened 2015-04-20
Add support of domain validation for "href" and "src" attributes

knowings:
Hello,

In the following PR, I'm proposing to support the ability of managing domain constraints on "href" and "src" attributes in the whitelist.

Basically, it is available like protocols. Developers can add or remove domains on a given tag and attributes.

Feel free to comments or suggests improvements and/or fixes.

Best regards.


jhy:
Thanks for this and sorry for the late reply. I haven't seen this as a commonly requested use-case, and so for the sake of simplicity I don't currently plan on incorporating it into jsoup. The implementation and test cases look good though. It would be interesting to see if we could a more general plugin mechanism to the cleaner to support custom validations.

PR #564, opened 2015-04-27
Add Element#getWholeText()

takezoe:
Add `Element#getWholeText()` to get raw text in the element easily.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -877,6 +877,27 @@ public class Element extends Node {
     }
 
     /**
+     * Get the (unencoded) text of all children of this element, including any newlines and spaces present in the original.
+     *
+     * @return text
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/org/jsoup/nodes/Element.java
```diff
@@ -877,6 +877,27 @@ public void tail(Node node, int depth) {
...
+                if (node instanceof TextNode) {
+                    TextNode textNode = (TextNode) node;
+                    accum.append(textNode.getWholeText());
+                }
+            }
+
+            public void tail(Node node, int depth) {
+            }
+        }).traverse(this);
+        return accum.toString().trim();
```
cuantico: why `trim()`? if we use `<p>     Hello\nthere     </p>`, wouldn't I want everything inside `<p>` that is not an html node?

jhy:
Thanks! Merged. I renamed it to #wholeText()

PR #565, opened 2015-04-28
Added limit(int) method

sravan953:
Hello,

As discussed, I've added the limit(int) method. Should be chainable.


jhy:
Thanks, but not seeing much value from this one-liner

PR #574, opened 2015-05-13
Use the XML parser if the MIME-type matches an XML type

hichris1234:
Currently, jsoup just uses the HTML parser unless otherwise specified. However, the MIME-type is checked to see if it's an XML, so to prevent issues with parsing, jsoup should use the XML parser when the MIME-type matches an XML type.


jhy:
Thanks. I merged this, after a few tweaks (docs, tests, and only switch if the user hadn't set the parser explicitly).

https://github.com/jhy/jsoup/commit/e6514a014b46ae76d5d3163c82f4f46003186197

hichris1234:
That's perfect. Thank you! :)

PR #582, opened 2015-05-27
Keep tag that have not been closed by > before reaching EOF.

fazerish:
We use Hibernate Validator (HV) and the @SafeHtlm annotation to validate input from users. During a security review we discovered that an unsafe XSS vector slipped by the validator. During debugging HV we discovered that the source of the problem was related to how Jsoup handled tags without a closing > when reaching EOF.

[Code reference to SafeHtmlValidator in HV](https://github.com/hibernate/hibernate-validator/blob/dfa65141d63cbfbb33f79522f6ba40a6d2b29b5c/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/SafeHtmlValidator.java#L75)

### DIFF
--- a/src/main/java/org/jsoup/parser/TokeniserState.java
+++ b/src/main/java/org/jsoup/parser/TokeniserState.java
@@ -757,6 +757,7 @@ enum TokeniserState {
                     break;
                 case eof:
                     t.eofError(this);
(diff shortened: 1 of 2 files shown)


jhy:
Looks good, thanks.

Issue #554, opened 2015-04-17
Unexpected behavior in elementSiblingIndex


DavidGauntt:
The documentation for elementSiblingIndex states "Get the list index of this element in its element sibling list. I.e. if this is the first element sibling, returns 0".

This would imply that if

    n=myElem.elementSiblingIndex();

then

    myElem.parent().children().get(n)==myElem.  

However, this is not how elementSiblingIndex behaves. What is guaranteed is that

    myElem.parent().children().get(n).equals(myElem).  

For example, if both row 2 and row 5 of a table are

    <tr><td>Cell1</td><td>Cell2</td></tr>

then the Element object associated with both rows will have the same `elementSiblingIndex()`.

jhy:
Is this the same issue as #561?

jhy:
Thanks for finding this, fixed

PR #587, opened 2015-06-10
fix issue 580

DarrelHsu:
```
System.out.println(Jsoup.parse("<html><body><table><tr><td> VALUE </td><td> VALUE </td></tr></table></body></html>").select("td"));
```


DarrelHsu:
this change will fix the issue 579 also

jhy:
This was already fixed with #554, right? Example now prints two values correctly, without this extra change.

Issue #552, opened 2015-04-13
Unexpected elements inside table are moved to wrong location


pfarrel:
This commit https://github.com/jhy/jsoup/commit/e99193605b688e923d20054c13db897cff751607 introduced a bug where handling of unexpected elements inside a table element changed, resulting in the unexpected elements being pushed further up the document than before.

I have constructed a minimal repro. Before the commit in question, the unexpected p tag would continue to be positioned after the comment (with some elements being closed early etc. to support this). After the commit, the p tag and its contents are moved up more than one table level, and now appear after the comment tag.

Obviously this input HTML is very broken and bad, but it seems that the change in behaviour was unintended. 

copying @jaredstehler  

    @Test
    public void testInvalidTableContents() throws IOException {
        File in = getFile("/htmltests/table-invalid-elements.html");
        Document doc = Jsoup.parse(in, "UTF-8");
        doc.outputSettings().prettyPrint(true);
        String rendered = doc.toString();
        int endOfEmail = rendered.indexOf("Comment");
        int guarantee = rendered.indexOf("Why am I here?");
        assertTrue("Comment not found", endOfEmail > -1);
        assertTrue("Search text not found", guarantee > -1);
        assertTrue("Search text did not come after comment", guarantee > endOfEmail);
    }

Uses the following fixture:

    <html>
        <body>
            <table>
                <tr>
                    <td>
                        <table>
                            <tr>
                                
                                <table>
                                    <p>Why am I here?</p>
                            </tr>
                        </table>
                    </td>
                </tr>
            </table>
        </body>
    </html>

jhy:
Thanks. I reviewed the commit and can't immediately see the problem or where it regressed.

jaredstehler:
So the issue seems to be that the pattern 
```
boolean processed = tb.processEndTag("table");
if (processed)
  return tb.process(t);
```

is broken with the change to reuse end tags via the ```processEndTag()``` method, in that it affects the tag referenced at ```t```.

For example, in handleMissingTr() (https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/parser/HtmlTreeBuilderState.java#L1135) :

```
        private boolean handleMissingTr(Token t, TreeBuilder tb) {
            boolean processed = tb.processEndTag("tr");
            if (processed)
                return tb.process(t);
            else
                return false;
```

[...]

PR #591, opened 2015-06-17
issue 552 fix

jaredstehler:
test case reproducing the element ordering issue, and a potential fix.

addresses https://github.com/jhy/jsoup/issues/552


jhy:
Thanks very much for your analysis and fix. I have implemented a slightly different fix, which is to test if the token about to be recycled is already in use. This should catch other instances of the same bug. I incorporated your test.

I did find another possible instance (in `<isindex>` normalization) which didn't actually cause bad output, but there would probably be other bugs lurking around there without this change.

jaredstehler:
excellent, thanks!

PR #597, opened 2015-07-08
Changed 'relative redirect with dot' test to show failing use case

xpadro:
Changed the test URL to show a use case where it fails to redirect to a URL which includes a relative dot.


jhy:
Thanks, seeing the host without a path showed me where the error was.

PR #609, opened 2015-08-07
Add mode to no escape entities and new setting to preserve attribute's quote

cuantico:
Two basic changes:
- Support the preserve of the quote used in any HTML Attribute, by enabling the outputsettings of "preserveQuote"
- Support the ability to not escape any HTML entity into its entity code. 

This is just trying to help keep the html code untouched when not expected. 

For example when HTML is used in combination with freemarker code (or custom tags) so you get some code inside attributes that you don't want escaped, or you don't want the quote of the attribute change from single quotes to double quotes because you have another code inside that has double quotes.

`<a href="${myURL!'default'}">test</a>`

### DIFF
--- a/src/main/java/org/jsoup/nodes/Attribute.java
+++ b/src/main/java/org/jsoup/nodes/Attribute.java
@@ -98,9 +98,16 @@ public class Attribute implements Map.Entry<String, String>, Cloneable  {
     protected void html(StringBuilder accum, Document.OutputSettings out) {
         accum.append(key);
         if (!shouldCollapseAttribute(out)) {
-            accum.append("=").append(quote);
+            accum.append("=");
+            if (out.preserveQuote()) 
+                accum.append(quote); 
+            else 
+                accum.append('"');
             Entities.escape(accum, value, out, true, false, false);
-            accum.append(quote);
+            if (out.preserveQuote()) 
+                accum.append(quote); 
+            else 
+                accum.append('"');
         }
     }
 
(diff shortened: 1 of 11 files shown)


remisbaima:
great feature!

mojavelinux:
> Support the ability to not escape any HTML entity into its entity code.

I agree, it would be nice to be able to output the document formatted without having the side effect of entities being modified. For example, if I have `&#8217;` in my input document, Jsoup either expands it to a Unicode glyph if the charset is UTF-8, or changes it to `&#rsquo;` or `&#x2019;` depending on the Entities.EscapeMode setting if the charset is ASCII. I would just like it to be as it was regardless of the charset.

jhy:
I'm never going to support an escape mode of none. It creates invalid or incorrect HTML. I.e. input of

```
<p>Hello &lt;script src="attacker"&gt;
```

With your change that would become after parsing and outputting:

```
<p>Hello <script src="attacker">
```

How is that valuable to anyone?

Maybe there's a case to be made for preserving a digit form vs hex or something (though I'm not sure what really - it's the same outcome when parsed).

As to preserving the quote I'm not really sure of the value of that either. Maybe there's something to be said for a serialization option which aims to minimize the output - so if able, don't use quotes, or use either double or single based on if the other is included (so no need to escape). But we're talking about trivial filesize differences with a reasonable amount of extra code complexity and either runtime complexity (scanning) or memory (in this PR, extra fields on every attribute to track)

Issue #614, opened 2015-08-24
Elements.select() does not return all matching Element objects


saibotsivad:
During an upgrade from 1.7.1 to 1.8.3 we have come across an issue that could be a bug, but also could be a misunderstanding.

### Setup test

To setup the test, we start with a simple HTML example:

```html
<!DOCTYPE html>
<html>
    <body>
        <div>word</div>
        <div>word</div>
    </body>
</html>
````

And then create en Elements object like this:

```java
Document document = Jsoup.parse("<!DOCTYPE html>...etc...</html>");
Elements elements = document.select("div");
```

### Unexpected/Inconsistent behaviour

If a `.select()` is run on the `elements` object, it only finds a single instance. E.g. this passes in 1.8.3 but not in 1.7.1:

```java
int elementsFoundUsingSelect = elements.select(":contains(word)").size();
Assert.assertEquals("should find two but finds one", 1, elementsFoundUsingSelect);
```

On the other hand, if I loop over the `elements` object and run a select on each `Element`, I get the following:

```java
int elementsFoundUsingSelect = elements.select(":contains(word)").size();
int elementsFoundUsingLoop = 0;
for (Element element : elements) {
	if (element.select("table:contains(word)").size() > 0) {
		elementsFoundUsingLoop++;
	}
}
Assert.assertEquals("should find two", 2, elementsFoundUsingLoop);
```

### Expected behaviour

I would expect `elements.select(":contains(word)")` to return two elements, but in 1.8.3 it only returns one element.

### Other notes

It appears that when the selected HTML content is identical (e.g. the two `<div>` elements in the above example are identical) than running a `.select()` on the `Elements` object will only select the first object.

### What do I do?

Which behaviour is correct? Am I misunderstanding the documentation?

* Previously we were relying on the 1.7.1 behaviour to find all instances of an identical element, and then navigate from *each* of those using `Elements.select()`
* However in 1.8.3 you can find all instances of identical elements, but you can't navigate from them using `Elements.select()` as it will only `select()` within the first one.

saibotsivad:
Here's [a gist](https://gist.github.com/saibotsivad/946916e2ce8bf30490ae) I made that might be easier to see.

saibotsivad:
[...]

Issue #561, opened 2015-04-21
changes made to isEqual on the Element now mess up the nextElementSibling / previousElementSibling / indexInList


andyphillips404:
The updates made between 1.8.1 and 1.8.2 in the isEqual of the Element now make the behavior of all functions utilizing the indexInList (which relies on the isEqual) to behave incorrectly.  The new isEqual only looks at the tag which then correctly gets the wrong element sometimes, in lists that  contain multiple items.

For example, multiple tr in tables, it can't find the current table index!   This shows up in the changes to the isEqual in 1.8.2.   I am more than happy to put a fix in, but i think we can either do two things:
1.  Change the isEqual back (but i'm not sure on the  motivation here why the change)
2.  Change the internal indexInList function in Element to not use isEqual but look for equal elemenet objects.

I am personally concerned that the change to isEqual is not robust enough check for isEqual as it appears to only check if that tag is equal....   But i also may not understand the full impact of the change and motivation

I'll put a pull request in, but was looking at feedback on which direction should be taken above!

jhy:
The motivation of the isEquals change is to make elements and their child trees comparable. Their nodes and attributes should be part of the equality check.

Can you start with a test case that displays the problem? From your description, 2 sounds like the approach to go with.

jhy:
You could think of the new behavior of equals to be like String.equals. I.e. check for same content with .equals. Check for identity with ==. Which is the bug, indexInList should be identity not contents.

jhy:
Thanks for reporting and the detail, fixed now

andyphillips404:
Great, thanks!

PR #616, opened 2015-08-25
Add breaking tests to test Element.equals

saibotsivad:
Related to #614, #561, and #554, this introduces two tests to assert that selectors return an expected number of elements.


saibotsivad:
Note that the failing checks are on purpose, pending a resolution to #614. I can make the code changes and/or update tests on resolution of #614.

jhy:
Thanks - fixed by changes for #614

Issue #535, opened 2015-03-11
selector "h3 a" not working properly


mrbrdo:
Hello,

According to documentation the selector "h3 a" should behave the same as a normal CSS selector.
However it is not working properly for me:

```scala
div.children.select("h3 a") // results: 0
div.children.select("h3").select("a") // results: 1
div.children.select("h3 > a") // => results: 1
```

These two should return the same results, but they don't. I have the same issue with other similar selectors.

Relevant part of source HTML:
```html
<div><h3 class="r" style="line-height:normal"><a class="l" href="http://sobe-macek.si/" onmousedown="return rwt(this,'','','','2','AFQjCNGL8bH57qrT59_z0XO7HGG6eyhJog','','0CEoQoAIwAQ','','',event)">Macek Rooms</a></h3></div>
```

If needed I can provide the full HTML file.

EDIT:
I realized that `div.select("h3 a")` also works properly. In some way I get it (if `children` is used, it will only search through children). However, this is really a mess now. The problem I have with `div.select` is that it will match the current element as well (which is unlike basically any other HTML parser that supports CSS selectors). I can't understand what you were thinking. It makes me feel like smashing my monitor or throwing my computer out of the window.
Anyway, rant off, but how am I then supposed to achieve the usual behavior, such as that of jQuery for example, that is for `el.select("h3 a")` to work while **not** matching `el` (in a general case)?

jhy:
Fixed by #619.

Issue #417, opened 2014-05-02
descend selector does not work for a special case


xzer:
Assume I have the following structure:

            <div>
              <span></span>
            </div>

Then, I have a node which represent the parent div node:

            Element divElement = doc.select("div").get(0);

Then, I perform a selector again:

            Elements foundElements = divElement.select("div span");

the result of foundElements is empty.

I digged the source and found the following source starting from row 59 of StructuralEvaluator:

            while (parent != root) {
                if (evaluator.matches(root, parent))
                    return true;
                parent = parent.parent();
            }

I believe that the right way of this loop should be as following:

            do {
                if (evaluator.matches(root, parent))
                    return true;
                parent = parent.parent();
            }while (parent != root);

jhy:
Fixed by #619

PR #619, opened 2015-08-30
Fix "Parent" Evaluator ignores "root" argument for matches.

asheiduk:
The "parent" Evaluator has a kind of "off by one" error - it does not match, when the expression requires matching the provided root element.

It seems to me, that this bugfix also solves  #417 and #535.


jhy:
Thanks! Merged

PR #620, opened 2015-08-31
Add Element#matches(cssQuery)

stipx:


stipx:
As the Element#is(query) method is now in, I'll close this one.

PR #623, opened 2015-09-03
Enable Attribute Value Validation

haggisandchips:
### Summary
This pull request adds validation for attribute values and refactors the existing protocol handling to make use of the new validation framework.

### Motivation
The last 2 projects that I have worked on have used Jsoup but also required me to perform further validation to restrict the content users could submit. For example in my current project we must permit the `onclick` attribute for `<a>` tags but we also have to support untrusted content so this clearly requires extra validation. This request provides a flexible framework to allow bespoke validation of attribute values and also provides two out-of-the-box validators for simple values and regular expressions.

The `onclick` use case clearly has anti-XSS uses but moreover I think this request is very much in keeping with the following motivation behind Jsoup as it provides much more control over attributes as well as elements:
> The cleaner is useful not only for avoiding XSS, but also in **limiting the range of elements the user can provide**: you may be OK with textual `a`, `strong` elements, but not structural `div` or `table` elements.

### Possible Improvement
On one of the projects I also implemented a specific style validator that parsed the semi-colon separated `style` value and cleaned/validated specific css styles. If you like and merge this request then I would work on that next.

### DIFF
--- a/src/main/java/org/jsoup/safety/BaseValueValidator.java
+++ b/src/main/java/org/jsoup/safety/BaseValueValidator.java
@@ -1,0 +1,14 @@
+package org.jsoup.safety;
+
+public abstract class BaseValueValidator implements ValueValidator {
+
+    private final boolean required;
+
+    protected BaseValueValidator(boolean required) {
+        this.required = required;
+    }
+
+    public boolean isRequired() {
+        return required;
+    }
+}
--- a/src/main/java/org/jsoup/safety/PatternValueValidator.java
+++ b/src/main/java/org/jsoup/safety/PatternValueValidator.java
@@ -1,0 +1,28 @@
+package org.jsoup.safety;
+
+import java.util.regex.Pattern;
+
+import org.jsoup.nodes.Attribute;
+import org.jsoup.nodes.Element;
+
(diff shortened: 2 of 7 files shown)


haggisandchips:
Actually something has just occurred to me. This patch does work ... but to the end user it would need an understanding of the inner workings. If a protocol is added then internally a required validator gets added which would be sufficient to pass the attribute - if the user then wanted to use the framework to also validate the target domain (for example) they would need to add that validator with required=true which is not obvious. And it also means they would be restricted to a single validator which would need to catch all safe values.

This behaviour is not the general case so it brings in inconsistencies when protocols are restricted.

I think it would be best to put the protocols implementation back the way it was and have the new validation framework sit on top and operate after those checks.

Before I do any further work on this I would like to hear your thoughts though.

Cheers
Ivor

haggisandchips:
Sorry for all the indecision but I've had a good think about this and I'm going to close this pull request then reopen it after reimplementing my solution with option 2 above - that is definitely the best fit with the existing design.

haggisandchips:
Ahhh! Can't reopen because I force pushed while PR was closed. Time to start again with a new PR.

PR #624, opened 2015-09-04
Implement attribute value validation

haggisandchips:
### Summary

This pull request enables validation of attribute values. In addition to allowing developer to create their own bespoke validators it also provides a couple of built-in validators: one for simple values and one for regular expressions.

_NOTE: This replaces pull request #623, sorry about the mess I made with that one ;-)_
### Motivation

The last 2 projects that I have worked on have used Jsoup but also required me to perform further validation to restrict the content users could submit. For example in my current project we must permit the `onclick` attribute for `<a>` tags but we also have to support untrusted content so this clearly requires extra validation. This request provides a flexible framework to allow bespoke validation of attribute values and also provides two out-of-the-box validators for simple values and regular expressions.

The `onclick` use case clearly has anti-XSS uses but moreover I think this request is very much in keeping with the following motivation behind Jsoup as it provides much more control over attributes as well as elements:

> The cleaner is useful not only for avoiding XSS, but also in **limiting the range of elements the user can provide**: you may be OK with textual `a`, `strong` elements, but not structural `div` or `table` elements.
### Possible Improvement

On one of the projects I also implemented a specific style validator that parsed the semi-colon separated `style` value and cleaned/validated specific css styles. If you like and merge this request then I would work on that next.

### DIFF
--- a/src/main/java/org/jsoup/safety/BaseValueValidator.java
+++ b/src/main/java/org/jsoup/safety/BaseValueValidator.java
@@ -1,0 +1,14 @@
+package org.jsoup.safety;
+
+public abstract class BaseValueValidator implements ValueValidator {
+
+    private final boolean required;
+
+    protected BaseValueValidator(boolean required) {
+        this.required = required;
+    }
+
+    public boolean isRequired() {
+        return required;
+    }
+}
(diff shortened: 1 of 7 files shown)


haggisandchips:
Bump.

Is there any interest in this pull request?

Issue #650, opened 2015-11-05
Cannot use single quotes in selectors


ipolevoy:
First, cool project!

Found a small issue: this HMTL:

```html
<body>
<p>This is <a data-file='123'>jsoup</a>.</p>
</body>
```

works with this selector: 

```
a[data-file="123"]
```

As you can see, the HTML has single quotes for `data-file` attribute, but the selector must use double quotes, and not working with single quotes: 

```
a[data-file='123']
```

This means that the selector inside Java code must use escaped double quotes: 

```java
Document doc = Jsoup.parse(responseContent());
Elements main = doc.select("a[data-file=\"123\"]");
```

This makes code look a bit dirtier than it should. Are there plans to support single quotes? 

By the way, I intend to use this in tests with http://javalite.io/testing, which looks like a perfect match.

derfloscher:
+1 thank you for the pull request

jhy:
My inclination would just be to use `a[data-file=123]`. The HTML using single- or double-quotes has nothing to do with it; in the DOM there's an attribute with key = `data-file` and value = `123`.

ipolevoy:
:+1:

ipolevoy:
If someone is interested, this is how JSoup is used in JavaLite: http://javalite.io/testing#testing-html-content

PR #655, opened 2015-11-15
#650 Cannot use single quotes in selectors

ipolevoy:
This small change makes it possible to use single quotes in queries like this one: 

```java
Document doc = Jsoup.parse(responseContent());
Elements main = doc.select("a[data-file='123']");
```
Before this fix, you could only use double quotes, even if single quotes were actually used in HTML.

### DIFF
--- a/src/main/java/org/jsoup/select/Evaluator.java
+++ b/src/main/java/org/jsoup/select/Evaluator.java
@@ -280,7 +280,8 @@ public abstract class Evaluator {
             Validate.notEmpty(value);
 
             this.key = key.trim().toLowerCase();
-            if (value.startsWith("\"") && value.endsWith("\"")) {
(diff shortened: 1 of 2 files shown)


ipolevoy:
Hey, Jonathan. 
Do you have plans for the next release? I wonder how fast this fix can find its way to Maven Central. 
I want to use your library, but this is a show stopper. 

Looking forward!
tx
Igor

cuantico:
Just curious, why do you need to use `div[class='main']` instead of `div[class=main]` ?

ipolevoy:
because this is now you normally write finders using JQuery, so consistency  across web development and tests makes sense. Besides  not sure if your approach will even work for attributes with multiple values:

```
div[class='main subheader']
```

ipolevoy:
:+1:

PR #662, opened 2015-12-02
Add appendTo method to Element.

sedran:
appendTo method appends the element to a given parent element and returns itself, instead of the parent. See jQuery's appendTo method for further information.
    As an example, use the code below;

```
element.appendTo(parentElement).attr('class', 'text-danger').text('Alert!!');
```

Instead of the code below;

```
element.attr('class', 'text-danger').text('Alert!!');
parentElement.appendChild(element);
```

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -291,6 +291,25 @@ public class Element extends Node {
     }
 
     /**
+	 * Add this node as a child node to the given parent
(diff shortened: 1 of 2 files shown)


jhy:
Thanks! Merged. Better late than never ;)

sedran:
Thank you, this is my first merged open source software contribution. :)

jhy:
Excellent, I hope it's the first of many!

PR #665, opened 2015-12-17
Modified to save memory when creating many unknown (custom) tags, plus option to format inline

gregko:
Currently a new Tag instance is created each time a tag with previously used tagName is needed, sometimes creating thousands of unnecessary Tag instance duplicates. This pull request proposes to fix this by maintaining a map of customTags, similar to the 'tags' map maintained for known tags, and returning a tag from this list, if the tagName is already there.

Also, for a custom tag that can "do anything", a choice is needed to decide if it should be formatted as block or not. For this the pull request provids a new variant of Tag valueOf(String tagName, boolean formatAsBlock).

LONG RATIONALE: custom, unknown tags may be used if the HTML and DOM are manipulated only for internal consumption in an app, and not for publishing on the web. For example, in my case I wrap words and sentences in HTML pages or long ebook chapters for the purpose of sending entire sentences to TTS engine for reading aloud with correct intonation, while being able also to highlight the currently read sentence. A long page or ebook chapter may have thousands of words, and I wrap them into &lt;w&gt; tags to have them short, instead of &lt;span class="wordFindSomeUniqueClassName"&gt;. I also need my words (and &lt;snt&gt; tags for sentences) formatted as inline elements, not blocks.

I also developed code for extracting and wrapping equivalents of JavaScript Range objects and "DocumentFragments". Maybe it could be part of jsoup as well - let me know if interested, see more at
 http://stackoverflow.com/questions/34301276/jsoup-equivalent-of-dom-range-operations-like-extractcontents-etc

This small pull request does not contain that code.

Greg Kochaniak, Hyperionics

### DIFF
--- a/src/main/java/org/jsoup/helper/HttpConnection.java
+++ b/src/main/java/org/jsoup/helper/HttpConnection.java
@@ -16,8 +16,12 @@ import java.net.URLEncoder;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.IllegalCharsetNameException;
+import java.security.GeneralSecurityException;
 import java.security.KeyManagementException;
+import java.security.KeyStore;
 import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateExpiredException;
 import java.security.cert.X509Certificate;
 import java.util.*;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


jhy:
Thanks for the PR. However by placing every unknown tag found into a static map, this would introduce a memory leak / DOS vector (i.e. deliberately or accidentally feeding a jsoup app HTML with random tags). It would be possibly OK to do it within the context of the current doc, but I wouldn't want to see it for a Parser (which can get reused) or for the whole lifecycle of the app.

Did you measure the memory benefit of this for your use case? Interested to see what the gain is.

Also why is block the only configurable option?

My inclination to make this a useable feature would be to create something like a ParseSettings object (like Document.OutputSettings) which is given at the Parser initiation time. That could be given a set of custom tags (and open up the control of block, inline, data, etc) which would override any canned ones. I was thinking of that settings object to control attribute normalization, etc.

Also, what's with the changes to HttpConnection? Seems irrelevant to the PR.

gregko:
Hello Jonathan,

I have no longer means of testing this, since some months ago I implemented my HTML DOM editing completely in C++, based on HTML Tidy parser and my own C++ code on top of it to do what I need (splitting text into paragraphs and sentences), it’s over 10x faster than Java. However in some long HTML texts (e.g. ebook chapters) I was creating thousands of custom tags <par> and <s> so the new tag names created for each instance were affecting performance and memory use.

 

The HttpConnection issue was later, I had trouble connecting to some secure web sites and needed to modify the way security certificates are created. I think it was an issue with connecting to web sites signed with a private security certificate, without the correct Certification Authority. Again, this was in February, so I don’t remember the details any more… I found some article on how Google wants security implemented now, added the code, tested – it worked, so I left it there and forgot. Sorry I’m so useless.

 

Greg

jhy:
Thanks again, but I won't take it in this current form. The idea is good and I'd like to see it in ParseSettings or similar to address my points.

PR #666, opened 2015-12-27
307 redirect should preserve request method & data

hatsuyuki15:
Currently the request method is changed to GET and the data is dropped when the request is redirected. 
However, according to https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection & https://tools.ietf.org/html/rfc7231#section-6.4.7, 307 redirect should preserve the original method and the data.

This patch corrects this problem.


jhy:
Thanks!

PR #672, opened 2016-01-28
Fix for bug 671: namespace is included in W3CDom.fromJsoup

ekasumov:
Added code to parse the namespace, if exists, and pass it to the NodeTraversor constructor and traverse() methods.  createElementNS() is used instead of createElement() when creating nodes in the W3C Dom document.
Added a unit test to validate that getNamespaceURI() and getLocalName() calls work on elements of the created W3C Dom document.
Added a sample XHTML file.


jhy:
Thanks, I merged this. I made some changes so that it supports multiple namespaces and tag prefixes.

PR #676, opened 2016-02-11
Added another list in Tag for tags that should be formatted as block.…

ridencww:
The select tag was pretty-printed as an inline element (select and all option tags on same line). Added a new list of tags that should be formatted as block elements and added select to the list. Wrote unit test to verify behavior.

### DIFF
--- a/src/main/java/org/jsoup/parser/Tag.java
+++ b/src/main/java/org/jsoup/parser/Tag.java
@@ -238,6 +238,9 @@ public class Tag {
             "meta", "link", "base", "frame", "img", "br", "wbr", "embed", "hr", "input", "keygen", "col", "command",
             "device", "area", "basefont", "bgsound", "menuitem", "param", "source", "track"
     };
+    private static final String[] formatAsBlockTags = {
+            "select"
+    };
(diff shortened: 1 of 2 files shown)


jhy:
Thanks. Why not just change it to a block tag?

Also your unit test doesn't verify your goal (that it gets pretty printed as a block tag).

ridencww:
I had considered moving SELECT to the block list, but I thought that creating a new list is a more targeted approach, complements the formatAsInline functionality, and potentially removes any unintended side effects from moving the tag from inline to block.  It seemed to me to be the lightest touch to the library.

I do see your point that I should specifically test the SELECT formatting.  My thinking regarding testing was that I needed to assert that the proper attributes were set for the SELECT tag based on the changes that I made and delegate asserting formatting based on what is returned by tag's formatAsBlock() to the existing tests. If you wish, I can add a test (in the ElementTest file?) to specifically assert proper formatting for SELECT. Just let me know.

jhy:
Thanks - on re-review - would prefer not creating the formatAsBlock method, particularly if there's only one thing in it.

PR #678, opened 2016-02-13
Overload hasClass method in Element.java to check for exact matches in class attribute

gnash48:
I have overloaded the hasClass() method in Element.java in order to allow client applications to test the class attribute of an element for an exact, case insensitive match. For example, if an HTML page has a tag:  `<p class="this is class"></p> ` , the client would have to implement the hasClass() three times as such: ` if(Element.hasClass("this") && Element.hasClass("is") && Element.hasClass("class")) ` , in order to acheive an exact match for the element's class attribute. With the method overloaded, the client can simply say:` if(Element.hasClass("this is class", true) `. Please excuse the useless deletion/additions on unaffected lines, my text editor had a file save bug. 

Note: The absoluteMatch argument is present simply to change the method signature of the new hasClass method

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -13,9 +13,9 @@ import java.util.regex.PatternSyntaxException;
 /**
  * A HTML element consists of a tag name, attributes, and child nodes (including text nodes and
  * other elements).
- * 
+ *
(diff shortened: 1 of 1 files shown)


jhy:
Thanks for the suggestion. However this breaks the meaning of a 'class'. Classes don't have spaces.

Also it's redundant, you could already just do `element.className().equals("this is a class")`. So why add another method.

PR #687, opened 2016-03-14
Added DetailedHtmlCleaner that validates HTML against a whitelist and…

makiftutuncu:
Added DetailedHtmlCleaner that validates HTML against a whitelist and returns a list of errors containing tags and attributes that were removed.


jhy:
Wouldn't it be better to add a tracer in the existing Cleaner? This looks duplicative and likely to get out of sync with the original.

mehmetakiftutuncu:
It definitely would. I'll work on it if I get free time. In the meantime, the idea is there if anybody wants to update the existing cleaner. :)

jhy:
Thanks -- closing out

PR #700, opened 2016-04-20
Exclude examples from artifacts

sebkur:
I noticed that the examples from the main source tree appear in the release artifacts. Assuming that this is not intended, I have setup a patch that prevents the classes from the org.jsoup.examples package from being included in the *.jar and *-sources.jar artifacts.

I think an alternative approach would be to move the examples to a separate source tree (or different project). What do you think?


jhy:
Thanks! Merged.

PR #702, opened 2016-04-25
add dataBinary method to send raw http data

ngocbd:
Add binaries data to request as same as CURL command `curl --data-binary URL `
It usedfull in many case e.g raw json , raw text (non key-value pair) , raw binary image/file see RawDataPost.java for example of use


jhy:
Thanks. I implemented basically the same thing with requestBody(String)

onriv:
But it seems still not very convenient for cases like posting with raw binary quest body.

eos1d3:
@jhy I do find many many times for uploading image with binary format, not in String, and Jsoup just does not work.

I do not see any way `requestBody(String)` can send binary as ByteArray. Pleases implement `dataBinary`. They are not the same thing.

PR #717, opened 2016-06-14
Text within an element to be queried without doing normalization as is

lakshmirajagopalan:
There were use-cases where we wanted text to be queried with normalizing the text inside (Not removing whitespace).

This seem to be a legit use-case with jsoup, there were lot of stackoverflow questions for doing the same

http://stackoverflow.com/questions/5454172/prevent-jsoup-from-discarding-extra-whitespace
http://stackoverflow.com/questions/20295098/how-to-parse-html-and-keep-all-line-breaks

If there is a better way to handle the same please let me know @jhy


jhy:
Thanks and sorry for the late reply.

https://jsoup.org/apidocs/org/jsoup/nodes/Element.html#wholeText() exists now for this use-case

Issue #723, opened 2016-06-28
XML Selector with any namespace


er1c:
I'm running into the same situation described here: http://stackoverflow.com/questions/23764992/parse-xml-nodes-having-text-with-any-namespace-using-jsoup

I would like to have an easy way to ignore the namespace when creating a selector.  I would probably propose: `*|E` that matches elements of type E in any namespace (wildcard) (https://jsoup.org/apidocs/org/jsoup/select/Selector.html)

jhy:
Thanks, I merged #724

PR #724, opened 2016-06-28
Add wildcard-namespace selector support '*|' Issue #723

er1c:
By using the "endsWith", it's possible an element named: "ns:foo:bar" with selector "*|bar" or "*|foo:bar", which shouldn't be valid (https://www.w3.org/TR/REC-xml-names/#ns-decl), but it's also fairly flexible/error-tolerant

### DIFF
--- a/src/main/java/org/jsoup/parser/TokenQueue.java
+++ b/src/main/java/org/jsoup/parser/TokenQueue.java
@@ -346,13 +346,13 @@ public class TokenQueue {
     }
     
     /**
-     * Consume a CSS element selector (tag name, but | instead of : for namespaces, to not conflict with :pseudo selects).
+     * Consume a CSS element selector (tag name, but | instead of : for namespaces (or *| for wildcard namespace), to not conflict with :pseudo selects).
      * 
      * @return tag name
      */
     public String consumeElementSelector() {
         int start = pos;
-        while (!isEmpty() && (matchesWord() || matchesAny('|', '_', '-')))
+        while (!isEmpty() && (matchesWord() || matchesAny("*|","|", "_", "-")))
             pos++;
         
         return queue.substring(start, pos);
--- a/src/main/java/org/jsoup/select/Evaluator.java
+++ b/src/main/java/org/jsoup/select/Evaluator.java
@@ -51,6 +51,28 @@ public abstract class Evaluator {
         }
     }
 
+
+    /**
+     * Evaluator for tag name that ends with
+     */
+    public static final class TagEndsWith extends Evaluator {
+        private String tagName;
+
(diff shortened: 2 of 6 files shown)


src/main/java/org/jsoup/parser/TokenQueue.java
```diff
@@ -346,13 +346,13 @@ public String consumeTagName() {
...
-     * Consume a CSS element selector (tag name, but | instead of : for namespaces, to not conflict with :pseudo selects).
+     * Consume a CSS element selector (tag name, but | instead of : for namespaces (or *| for wildcard namespace), to not conflict with :pseudo selects).
      * 
      * @return tag name
      */
     public String consumeElementSelector() {
         int start = pos;
-        while (!isEmpty() && (matchesWord() || matchesAny('|', '_', '-')))
+        while (!isEmpty() && (matchesWord() || matchesAny("*|","|", "_", "-")))
             pos++;
```
joshgit: Do you want += 2 for "*|" ?

src/main/java/org/jsoup/select/Evaluator.java
```diff
@@ -51,6 +51,28 @@ public String toString() {
...
+        }
+
+        @Override
+        public boolean matches(Element root, Element element) {
+            return (element.tagName().endsWith(tagName));
+        }
+
+        @Override
+        public String toString() {
+            return String.format("%s", tagName);
```
joshgit: Shouldn't this just return tagName?

src/main/java/org/jsoup/parser/TokenQueue.java
er1c: it technically works by coincidence, since the 2nd time around the "|" will also match, I was considering just adding the `'*'` as a character in the list, but did want to force the combined `*|` heuristic to match.  I'm not sure if something like: `while (!isEmpty() && (matchesWord() || matchesAny('|', '_', '-') || (matches("*|") && (pos++ > 0)) pos++;` is better, or just more confusing from a code readability standpoint.  Open to thoughts/suggestions @jhy

src/main/java/org/jsoup/select/Evaluator.java
er1c: Following the convention for the Tag Evaluator: https://github.com/er1c/jsoup/blob/5d00201aec621e81a37447e32baf1a82040ee8e4/src/main/java/org/jsoup/select/Evaluator.java#L49  I believe this technically handles "null" more appropriately for display purposes:

```
java> String.format("%s", null)
java.lang.String res0 = "null"
```

er1c:
@jhy any chance for a review?

jhy:
Looks great, thanks!

PR #725, opened 2016-06-29
Add Connection.headers(map)

zJiaJun:
Add Connection headers method(key-value pair)
It usedfull in many case


zJiaJun:
Hi, Jonathan. 
Do you have plans for the next release?

jhy:
Thanks!

PR #734, opened 2016-07-16
Allow setting custom content-type with POST request

### LINKED ISSUE #627: cannot set content type in the header
Connection.Response res = Jsoup.connect(url) 
					.header("Content-Type", "application/json")
					.method(Connection.Method.POST)
					.execute();

the content type still is the default value (application/x-www-form-urlencoded).

hatsuyuki15:
Fix #627 
Currenty, regardless of whether the user has set the content-type header or not, it will always be changed to either 'multipart' or 'x-www-form-urlencoded' with POST request. 
This pull will allow user to set a custom content-type for POST request. If the header is not set, then change it to either 'multipart' or 'x-www-form-urlencoded' as before.


ArloL:
I believe this has been fixed by https://github.com/jhy/jsoup/commit/84ffc4b73ca48ebe2423eac7a5f7b0aaa6c93aad

jhy:
Thanks @hatsuyuki15  and @ArloL, yes that commit reached the same goal.

PR #753, opened 2016-08-23
optimize Element#hasClass

### LINKED ISSUE #752: Optimize Element#hasClass
Profiling shows that this method is really hot (CPU and memory allocation).
The regexp split should be replaced by a custom solution that do not allocate unneeded memory.
the parsing of the attribute should also be stopped if there is a match.

benbenw:
this method is perf sensitive (CPU and memoryallocations)
This replaces the regexp split with a custom implementation that do not
allocate unneeded memory.
With this patch a simple test calling Element#select in a loop for
several selector went from 3Gb allocated to 10Mb.
GC times were reduced.
Fixes #752

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -1067,15 +1067,49 @@ public class Element extends Node {
     */
     public boolean hasClass(String className) {
         String classAttr = attributes.get("class");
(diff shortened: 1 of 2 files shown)


jhy:
Thanks, looks like a good approach.

jhy:
Thanks, merged

PR #755, opened 2016-08-23
Optimize Attribute#iterator

### LINKED ISSUE #754: Optimize Attribute#iterator
It allocates a lot of memory during the parsing through ParseSettings#normalizeAttributes.

benbenw:
It allocates a lot of memory during the parsing through
ParseSettings#normalizeAttributes.
this new implementation just wrap the LinkedHashMap values iterator in a
read-only iterator.

Iterating 1000 times over an internal html file (~800Kb) was allocating
970Mb, with this optimization it's now 350Mb

Fixes #754

### DIFF
--- a/src/main/java/org/jsoup/nodes/Attributes.java
+++ b/src/main/java/org/jsoup/nodes/Attributes.java
@@ -172,7 +172,11 @@ public class Attributes implements Iterable<Attribute>, Cloneable {
     }
 
     public Iterator<Attribute> iterator() {
-        return asList().iterator();
+        if (attributes == null || attributes.isEmpty()) {
+            return Collections.<Attribute>emptyList().iterator();
+        }
+        
+        return new AttributesReadOnlyIterator(attributes.values().iterator());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jhy:
Wha

jhy:
(Sorry, clicked wrong button) - looks good, just wanted to check what code coverage exists for this iterator or if we need another test?

benbenw:
Add a new commit with some tests of Optimize Attribute#iterator.

pmouawad:
@jhy ,any plan to release a jsoup version soon ?

Thanks for your nice library and your work.
Regards

jhy:
What do you think about just returning the LinkedHashMap iterator directly? Instead of wrapping it in a new object? Saves some GC, and allows a bit of extra functionality (like remove).

benbenw:
I tried to keep the same 'api' : in the current version it's not possible to remove an element from the iterator.
I'm afraid most of the remaining memory is allocated from the LinkedHashMap#values#iterator creation (should be verified), but removing an extra layer is always good.

jhy:
I think it's OK to change the API as it's backwards compatible and we didn't make a big point in the doc or anything that it was an immutable iterator. As you say it's not going to be a big improvement but 16 bytes is 16 bytes.

jhy:
Thanks for this, merged

PR #760, opened 2016-09-23
Change OuterHtmlVisitor visibility, reduce stream reading on startup

kirillF:
Move xhtml and base html charsets to memory from .properties to reduce stream reading on startup (Android sensitive). Change OuterHtmlVisitor visibility to public. Helps to modify nodes and construct outer HTML in single tree traverse.


jhy:
Hi - were you able to measure an impact from this? My thought is that the full entities are still loaded on init and that's where most of the time is going to be spent (i.e. only moving the base entities out doesn't seem like it would be impactful).

kirillF:
Hi, yes, full entities are still loaded on init and impact is not recognizable. Second thing is generating changing visibility of OuterHtmlVisitor allows to traverse document tree, update/format/etc and construct outer html in one traverse.

jhy:
I refactored Entities to load faster. Thanks.

PR #766, opened 2016-10-06
README changes

zygimantus:
Applied markdown for README file and Travis build status added.


jhy:
No thanks, they don't need to get a ping on every project view here. Anyway, it's not like the build is unstable.

PR #772, opened 2016-10-20
Maven plugins updated to latest versions

offa:
Maven plugins updated to latest versions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
   		<name>The MIT License</name>
   		<url>https://jsoup.org/license</url>
   		<distribution>repo</distribution>
-  	</license>  
+  	</license>
   </licenses>
   <scm>
   	<url>https://github.com/jhy/jsoup</url>
@@ -36,7 +36,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>2.0.2</version>
+        <version>3.5.1</version>
         <configuration>
           <source>1.5</source>
           <target>1.5</target>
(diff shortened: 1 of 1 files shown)


jhy:
Mmm I seem to remember hitting a whole bunch of problems with the Sonatype publish steps when I tried pushing these up before, and had to weave a specific set of versions. Will need to be careful with this. Thanks though!

offa:
Are there any ways to test without killing something?

jhy:
Yes, I will try a snapshot publish before merging.

jhy:
Thanks, looks like it worked fine. I hit some issues finding the updated command line args for doing the deployment, but not with any of the versions specifically.

offa:
Perfect!

jhy:
But it looks like Travis is having a fitz on OpenJDK6 where one of the plugins needs a higher lang version?

offa:
hmm, according to the [Apache Felix FAQ](https://felix.apache.org/documentation/faqs/apache-felix-bundle-plugin-faq.html#why-do-i-get-an-exception-unsupported-majorminor-version-490-when-i-build-with-a-14-or-earlier-jdk) everything >= JDK5 is good; but even earlier versions are still possible.

offa:
I'll open an PR with a workaround (downgrade _maven-bundle-plugin_) in a moment.

offa:
The [Dependencies List](http://felix.apache.org/components/bundle-plugin/dependencies.html) doesn't include any JDK7 related things. I've tested with a cleaned travis cache, but same problem.

offa:
Reason found: http://maven.apache.org/docs/history.html

Since maven v3.3.1 JDK7 is required. Last JDK5: 3.1.1, JDK6: 3.2.5

PR #776, opened 2016-10-23
Workaround for #772: Downgrade maven-bundle-plugin to v2.x.

offa:
Workaround for #772: Downgrade **maven-bundle-plugin** to **v2.x**. This enables JDK6 builds.


jhy:
Great, thank you!

PR #783, opened 2016-11-03
Update Document.java

nirro01:
cosmetic change in createShell


jhy:
Thanks - but a bit trivial! :)

Issue #791, opened 2016-11-23
Error when set Content-Type to "application/json"


dpy1123:
Hi there,
    I am using `Jsoup.connect(URL).requestBody(json).header("Content-Type", "application/json").post()` as mentioned in API DOC.
but get error `Unhandled content type. Must be text/*, application/xml, or application/xhtml+xml`

I set `connect.ignoreContentType(true)` to avoid this error, but I hope you can add this type in your code, and do not throw a `UnsupportedMimeTypeException`.

Since this is a very little issue, I don`t want to send a pull request.
 Thx~

jhy:
Thanks, but not thanks. json is not HTML and Jsoup should throw an exception. The default case is that people want to parse HTML out of the connection. If they just want the body string, that's what the purpose of ignoreContentType is. Otherwise what's the argument for putting json in there but not any/every other mime type?

PR #792, opened 2016-11-23
Update HttpConnection.java

dpy1123:
fix for #791


jhy:
Thanks, but not thanks. json is not HTML and Jsoup should throw an exception. The default case is that people want to parse HTML out of the connection. If they just want the body string, that's what the purpose of ignoreContentType is. Otherwise what's the argument for putting json in there but not any/every other mime type?

PR #800, opened 2016-12-15
return body content when it is error

cpusoft:
some web server will return error reason in body content.  
So , get the body content when is error, and put it into the HttpStatusException to return


src/main/java/org/jsoup/HttpStatusException.java
```diff
@@ -8,11 +8,13 @@
 public class HttpStatusException extends IOException {
     private int statusCode;
     private String url;
-
-    public HttpStatusException(String message, int statusCode, String url) {
```
jhy: Breaks API contract. You could add a new constructor

src/main/java/org/jsoup/HttpStatusException.java
```diff
@@ -8,11 +8,13 @@
...
+    private String errorDetail;
+    
+    public HttpStatusException(String message, int statusCode, String url,String errorDetail) {
```
jhy: Formatting issues throughout

src/main/java/org/jsoup/helper/HttpConnection.java
```diff
@@ -675,9 +675,25 @@ else if (methodHasBody)
                     }
                     return execute(req, res);
                 }
-                if ((status < 200 || status >= 400) && !req.ignoreHttpErrors())
-                        throw new HttpStatusException("HTTP error fetching URL", status, req.url().toString());
-
+                if ((status < 200 || status >= 400) && !req.ignoreHttpErrors()){
```
jhy: This is duplicative, look to extract into a common method

jhy:
(Closing out -- would be happy to review again if you address my comments)

PR #805, opened 2017-01-05
Return an empty Elements object

fr33d00m:
I've gotten a null pointer exception on an element by calling children() method.

I've fixed this cornercase with a guard-statement to return a new Elements object if childNodes are null.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -220,9 +220,7 @@ public class Element extends Node {
      * @see #childNodes()
      */
     public Elements children() {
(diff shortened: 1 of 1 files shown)


jhy:
Thanks, but how are you getting an NPE here in the first place? The constructor starts by initialization childNodes to be the empty list. childNodes should never be null and if there is a path to setting it to null I would prefer to guard that than every reader.

I added a test case to validate that elements with no child elements currently return the empty list - d8eb9bd63c861132e5307a65e8f2f234fab2416b - that passes on the current impl

fr33d00m:
I also cannot reproduce, thank you for your valuable time.

PR #813, opened 2017-01-18
Update Whitelist.java

springrain:
// it is bug, eq: \ ,protocol is "/", not "/:"
 //prot += ":";

### DIFF
--- a/src/main/java/org/jsoup/safety/Whitelist.java
+++ b/src/main/java/org/jsoup/safety/Whitelist.java
@@ -540,7 +540,8 @@ public class Whitelist {
                 }
             }
 
-            prot += ":";
+           // it is bug, eq: <img src="/springrain/images/test.jpg" /> ,protocol is "/", not "/:"
+           //prot += ":";
 
             if (value.toLowerCase().startsWith(prot)) {
                 return true;
--- a/src/main/java/org/jsoup/safety/Whitelist.java
+++ b/src/main/java/org/jsoup/safety/Whitelist.java
@@ -540,8 +540,10 @@ public class Whitelist {
                 }
             }
 
-           // it is bug, eq: <img src="/springrain/images/test.jpg" /> ,protocol is "/", not "/:"
-           //prot += ":";
+            // it is bug,  eq: \<img src="/springrain/images/test.jpg" /> ,protocol is "/", not "/:"
+            if(!"/".equals(prot)){
+               prot += ":";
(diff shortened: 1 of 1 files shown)


springrain:
// it is bug,  eq: \ ,protocol is "/", not "/:"
            if(!"/".equals(prot)){
               prot += ":";
            }

jhy:
Can you add a test case that demonstrates this (should fail without your patch)? I'm not sure what you're saying the bug is (because that looks like a relative path which would already be validated OK with a base URL or relative links enabled).

Also can you clean up the comments so it's ready for committing?

springrain:
package org.springrain.frame.util;

    import org.jsoup.Jsoup;
    import org.jsoup.safety.Whitelist;

    public class JsoupUtils {
	
		Whitelist user_content_filter = Whitelist.relaxed();
		user_content_filter.addProtocols("img", "src", "/");
               //Rich text editor to submit the image path, which is able to display normal, belonging to the normal path
	         String str="";
		//Clean up XSS
	       String input = Jsoup.clean(str, user_content_filter);
		//I want to print "" and actually print ""
		System.out.println(input); 
		
	       }
	
    }

jhy:
Right, so what you actually want to do is preserve relative links. That's already supported with the [Whitelist#preserveRelativeLinks(true) ](https://jsoup.org/apidocs/org/jsoup/safety/Whitelist.html#preserveRelativeLinks-boolean-)option.

Here's an example:

```java
@Test public void relative() {
  String html = "<img src='/springrain/images/test.jpg' />";
  Document unsafe = Jsoup.parse(html, "http://example.com/");
  Whitelist whitelist = Whitelist.relaxed().preserveRelativeLinks(true);
  String cleanBody = new Cleaner(whitelist).clean(unsafe).body().html();

  assertEquals("<img src=\"/springrain/images/test.jpg\">", cleanBody);
}
```

I'll close this pull request. Thanks.

jhy:
Oh and here's a simpler example that's the actual test case:

```java
@Test public void resolvesRelativeLinks() {
    String html = "<a href='/foo'>Link</a><img src='/bar'>";
    String clean = Jsoup.clean(html, "http://example.com/", Whitelist.basicWithImages());
    assertEquals("<a href=\"http://example.com/foo\" rel=\"nofollow\">Link</a>\n<img src=\"http://example.com/bar\">", clean);
}
```

PR #817, opened 2017-01-25
Minor fix for the parsing of non-empty selects

lexspoon:


jhy:
Was the bug that it was emitting a parse error whenever there was a start option tag?

lexspoon:
@jhy yes, exactly. I have a test here that runs a generated HTML file through Jsoup and insists that there are no errors.

jhy:
Thanks!

Issue #830, opened 2017-02-19
".class" selector seems to be failing


edasaki:
A more detailed post: http://stackoverflow.com/questions/42322374/jsoup-basic-class-css-selectors-not-working-html-shows-elements-exist-an

The problem can be summarized like so:

Debugging code:

    Elements as = doc.select("a");
    for (Element e : as) {
        if (e.classNames().size() > 0)
            M.debug(e.classNames());
        if(e.classNames().contains("bigChar")) {
            M.debug("found!");
        }
    }
    as = doc.select(".bigChar");
    M.debug("now: " + as);
    for (Element e : as) {
        if (e.classNames().size() > 0)
            M.debug(e.classNames());
    }

Output:

    [logo]
    [bigChar]
    found!
    [dotUnder]
    [dotUnder]
    [dotUnder]
    [dotUnder]
    [dotUnder]
    now: 

I've spent a while debugging this and I think the problem may have to do with malformed HTML. Attached is a sample input, and below is some simple code to run it (I am able to reproduce this problem consistently by reading from the file):

        try {
            Document doc = Jsoup.parse(new File("blahblah.txt"), "UTF-8");
            System.out.println(doc.html());
            System.out.println("FINAL");
            System.out.println(doc.select(".bigChar"));
            System.exit(0);
        } catch (IOException e) {
            e.printStackTrace();
        }

[blahblah.txt](https://github.com/jhy/jsoup/files/785471/blahblah.txt)

You can see in the HTML output that the "bigChar" class is used, but the .bigChar selector doesn't seem to find it.

In my run of the above code, my output is 

```
   </div> 
   <script type="text/javascript">
         $('#footer').css('top', $(document).height() - $('#containerRoot').height());
     </script>   
  </body>
 </html>
 FINAL
 

```

(i.e. nothing found for .bigChar)

edasaki:
Further debugging on the same blahblah.txt input:


            for (Element e : doc.getAllElements()) {
                if (e.hasClass("bigChar")) {
                    M.debug("Found iteratively");
                } else {
                    if (e.className().length() > 0)
                        M.debug(e.className());
                }
            }

Upon outputting all className()s in the Document, the bigChar is found in a className():
```
barTitle
barContent
arrow-general
bigChar
info
info
dotUnder
info
```

However, "Found iteratively" is never printed, so hasClass() is not finding it. 

Adding `e.className().contains("bigChar")` to the same if statement, like so, does find it.

[...]

PR #831, opened 2017-02-19
Revert potentially bugged Element.hasClass()

edasaki:
See issue #830 

If it turns out I'm totally wrong, please disregard this pull request. It seems to be the source of the problem, although unfortunately I don't have time at the moment to test it as thoroughly as I'd like.


jhy:
Thanks. As mentioned, it's fixed in 1.10.3

PR #862, opened 2017-04-07
Speed up matching CSS selectors and other tasks which use elementSiblingIndex() see #863

LukeButtersFunnelback:
Matching css selectors can be rather slow because finding the elementSiblingIndex() is relatively slow. In this I modify Element to cache a Map of its child elements to their elementSiblingIndex(). This results in a massive speed up. Checking a parsed CSS selector against 10.8M elements the time goes from 38.5s to 0.8s. The CSS selector I am using is `html > body > blockquote:nth-child(1209) > a:nth-child(11)`

The change involves:
1. Creating a ChangeNotifyingList, which wraps a given List and will call the given Runnable each time the List is modified.
2. Change childNodes in Node to use a ChangeNotifyingList.
   * We are able to follow the previous behavior of using a static final EMPTY_LIST to avoid creating unnecessary objects when a Element is created.
3. When childNodes in Node is modified the ChangeNotifyingList calls Node#onChildNodeChange()
4. Element is updated to have a field childElementToSiblingElementIndex, which maps its child Elements to their sibling element index.
   * The field is initially null and is set the first time a element tries to find its sibling element index.
   * The field is set back to null any time the childNodes list is modified.

This should result in Jsoup being no slower to parse Documents or edit documents, while dramatically speeding up CSS selector matching. Further with the ChangeNotifyingList future developers are free to modify the childNodes list without needing to add special code to null the Element.childElementToSiblingElementIndex field.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -1,5 +1,18 @@
 package org.jsoup.nodes;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
 import org.jsoup.helper.StringUtil;
 import org.jsoup.helper.Validate;
 import org.jsoup.parser.ParseSettings;
@@ -13,18 +26,6 @@ import org.jsoup.select.NodeVisitor;
 import org.jsoup.select.QueryParser;
 import org.jsoup.select.Selector;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.regex.Pattern;
-import java.util.regex.PatternSyntaxException;
-
 /**
  * A HTML element consists of a tag name, attributes, and child nodes (including text nodes and
  * other elements).
@@ -37,6 +38,14 @@ public class Element extends Node {
     private Tag tag;
 
     private static final Pattern classSplit = Pattern.compile("\\s+");
+    
+    /** 
+     * Maps child elements to their sibling element index
+     * 
+     * This is set to null when the childNodes List is modified, and is 
+     * re-coputed when a child finds out its sibling element index. 
+     */
+    private Map<Element, Integer> childElementToSiblingElementIndex = null;
 
     /**
      * Create a new, standalone element.
@@ -226,6 +235,7 @@ public class Element extends Node {
             if (node instanceof Element)
                 elements.add((Element) node);
         }
+        
         return new Elements(elements);
     }
 
@@ -614,8 +624,31 @@ public class Element extends Node {
      */
     public Integer elementSiblingIndex() {
        if (parent() == null) return 0;
-       return indexInList(this, parent().children()); 
+       return parent().getSiblingIndexOfChild(this);
     }
+    
+    /**
+     * Get the list index of the child element in its element sibling list. I.e. if this is the first element
+     * sibling, returns 0.
+     * 
+     * @param child The child element of this to find the sibling index of.
+     * @return child's position in element sibling list.
+     */
+    private Integer getSiblingIndexOfChild(Element child) {
+        // If the childElementToSiblingElementIndex is null we need to 
+        // re-compute the sibling indexes. It may be set back to null when
+        // the childNodes list is modified.
+        if(this.childElementToSiblingElementIndex == null) {
+            this.childElementToSiblingElementIndex = new HashMap<Element, Integer>();
+            List<Element> children = this.children();
+            for(int i = 0; i < children.size(); i++) {
+                this.childElementToSiblingElementIndex.put(children.get(i), i);
+            }
+        }
+        return this.childElementToSiblingElementIndex.get(child);
+    }
+    
+    
 
     /**
      * Gets the last element sibling of this element
@@ -1286,5 +1319,13 @@ public class Element extends Node {
     @Override
     public Element clone() {
         return (Element) super.clone();
+    }
+    
+    @Override
+    protected void onChildNodeChange() {
+        // The list of child nodes has changed, set the child element to index 
+        // map to null to force it to be re-computed if needed.
+        this.childElementToSiblingElementIndex = null;
+        super.onChildNodeChange();
     }
 }
--- a/src/main/java/org/jsoup/helper/ChangeNotifyingList.java
+++ b/src/main/java/org/jsoup/helper/ChangeNotifyingList.java
@@ -1,0 +1,148 @@
+package org.jsoup.helper;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+
+/**
+ * A list wrapper which can notify a runnable when the list is modified.
+ *
+ * @param <E>
+ */
+public class ChangeNotifyingList<E> implements List<E> {
+
+    /** Delegate list where changes will be made to.*/
+    private List<E> delegateList;
+    
+    /** Runnable to be called when the delegate list is modified */
+    private final Runnable onChange;
+    
+    /**
+     * 
+     * @param list List to wrap. This list should only be modified via this wrapper class.
+     * @param onChange The runnable to call when the wrapped list is modified by this.
+     */
+    public ChangeNotifyingList(List<E> list, Runnable onChange) {
+        this.delegateList = list;
+        this.onChange = onChange;
+    }
+
+    /**
+     * Called when a changing operation is made to the delegate list. 
+     */
+    private void changeHappend() {
+        this.onChange.run();
+    }
+
+    public int size() {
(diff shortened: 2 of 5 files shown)


jhy:
Thanks @LukeButtersFunnelback, looks like an interesting approach.

My first thought is to how this effects memory use and GC, particularly on Android. I am concerned with adding a Runnable and a Map (or slots for them) into every Node. Particularly as this will impact every user, but I think your case (select against 10M nodes by index) is a bit of an edge case.

Also the ChangeNotifyingList / Runnable seems to add a bit of indirection complexity but is only used for this one case; it's not something that we're leveraging elsewhere to make value of the added code.

What if we tracked just the index cache during a Selector that is heavily using indexes? Have you profiled and see that it is looping heavily in finding the child index during a select?

src/main/java/org/jsoup/nodes/Element.java
```diff
@@ -37,6 +38,14 @@
     private Tag tag;
 
     private static final Pattern classSplit = Pattern.compile("\\s+");
+    
+    /** 
+     * Maps child elements to their sibling element index
+     * 
+     * This is set to null when the childNodes List is modified, and is 
+     * re-coputed when a child finds out its sibling element index.
```
LukeButtersFunnelback: note to self spelling coputed -> computed

LukeButtersFunnelback:
Hi thanks for replying!!!!
* I am not sure how jsoup is typically used but for me I may have `n` CSS selectors and `m` documents to match them against. But actually because of the design I need to attempt each selector against every element, using `element#is`. I think other people have had issues about CSS selector speeds.
* Ah memory usage something i did not think about. To test this i just parsed this play `http://shakespeare.mit.edu/troilus_cressida/full.html` and kept adding the `Document` to a list until a OOM occurred. I set 256MB of memory. Here are the results:
** `152` documents with jsoup as is
** `139`documents with these changes.
** `149` documents if I remove the runnable and alter the `ChangeNotifyingList` to extend `ArrayList` (less fields on the classes).
** `121` with the above change + calling `elementSiblingIndex()` on every element (this creates that HashMap)
** `145` with the above change to the `ChangeNotifyingList` and runnable + wrapping the Map in a WeakReference.
* So yes it seems that this uses more memory less than 1% more but still more. Is that too much?
* Memory in terms of GC, this is a hard one, I guess more GCing is being done because of the extra map, however I could break evenish if I don't call `org.jsoup.nodes.Element.children()` which creates a new array list.
** Actually I notice other methods call that `children()` method but then do not use the new returned array they just want to do something on each element. I guess this is a similar wast of GC resources perhaps if the GC burden from the extra WeakRefernce<Map> is a problem then the `children()` method must also be burden that we should do something about it (in another MR). (See below about some util methods I wrote to avoid creating this array).
* The speed up right now is only for `org.jsoup.nodes.Element.getSiblingIndexOfChild(Element)`. The `org.jsoup.select.Evaluator.IsNthChild` class is the evaluator that happens to be using it for my selector.
* I guess this could be used in `org.jsoup.nodes.Element.nextElementSibling()` as well as `org.jsoup.nodes.Element.previousElementSibling()`
** I happen to know of those because I wrote some Jsoup helper methods to NOT call them and use my own faster methods :(
** I wanted to avoid doing as many changes as possible in this MR, hence I didn't alter all the methods to use this.
* Right now I don't know if it is a good idea to attempt caching of element sibling indexes outside of the Element/Node classes.

Now for some pictures of sampling with jvisualvm.
Here is the code:
```
        String css = "html > body > blockquote:nth-child(1209) > a:nth-child(11)";
        Evaluator parsedCss = QueryParser.parse(css);
        for(int j = 0; j<100000;j++) {
            for(Element e : doc.getAllElements()) {
                e.is(parsedCss);
            }
        }
```
Before:
[Before](http://i.imgur.com/GgY4xFd.png)
With the changes that are currently in the MR
[After](http://i.imgur.com/ew3NKyv.png)

I guess at this point I would like to know if this is change that you would be willing to accept or what needs to be done for you to be happy with it?

jhy:
Thanks very much for the depth of analysis and detail here @LukeButtersFunnelback, this is helpful.

I do like the thrust of the change. I'm just trying to work through what other approaches there may be. Logically this is a similar approach to the concept I mentioned in `Element.children()`:

> create on the fly rather than maintaining two lists. if gets slow, memoize, and mark dirty on change

It's interesting in your benchmark profile that the self time is spent in `children()`, not in `indexInList()`. That indicates to me that the time is spent in creating the Element List array, not in the == search.

I think we may be able to get effectively the same performance gain with less code if we hold onto the array create in children, and set it to be null when reindexChildren() is called. That's similar to this Map approach but should reduce some field redundancy and keep everything using the same path. I'll try that in a bit.

Also, we can probably get a speed improvement by changing the `Evaluator.IsNthChild` to grab the element by index and see if it is equals, vs the `calculatePosition` implementation there is now.

Also, separately, I'm interested in why you're looping on `Element.is()`. Why not call `document.select()` with the same query and get that list?

LukeButtersFunnelback:
Hi

I did switch to what children() is doing to not return an array and that does speed things up but not as significantly as the cached map change did. Smashing out such a method I have this code:
```
    public Integer elementSiblingIndex() {
       if (parent() == null) return 0;
       
       
       return parent().getSiblingIndexOfChildNoArray(this);
       //return parent().getSiblingIndexOfChild(this);
    }
    
    private Integer getSiblingIndexOfChildNoArray(Element child) {
        int index = 0;
        for (Node node : childNodes) {
            if (node instanceof Element) {
                if(node == child) {
                    return index;
                }
                index++;
            }
        }
        return null;
    }
```
And the timings become:
* `3960ms` no changes
* `161ms` with the cache Map
* `1060ms` without the creation of the Array list ie with the code above.
[No array creation](http://i.imgur.com/vtNufE8.png)
As you can see time is now spent in `getSiblingIndexOfChildNoArray()` which is to do with the `instance` and `==` checking.


hmm I never did try it that way as  it wasn't simple to do in the layout of the code which is very `onElement` focused. So I had a go (Note i used `Selector.select(parsedCss, doc)` to pre-parse the css) and it seemed to take `3754ms` here is what it looks link under the profiler:

[select()](http://i.imgur.com/QNCzckC.png)

still plenty of time spent in that `Element.children()` method.

jhy:
Here's a draft of what I was thinking - caching (memoizing) the child element array after it's done the filter, and clearing it on changes: baffcb2c573fd937b72843eeda9e8cc0559c429d

On my laptop, with 1000 iterations (of the play and selector in your example) it was taking 59 seconds originally, and now is at 3 seconds. That's a bit slower than your impl but still a great saving, and I think the code is a bit simpler, creates a little less GC (which I haven't profiled on Art yet), and it hits the other methods as well so they will get the same improvement.

The next step would be to improve the Evaluator so it pulls by index vs looping and evaluating which will be a good speed bump.

LukeButtersFunnelback:
Hi I tried that out ant is takes `300ms` for 100 iterations for me vs `178ms` for the map. If you really prefer this approach I would be happy with that (10 times faster is probably fast enough).
I think though you may want to:
* using something similar to my ChangeNotifyingList except to save memory you make it abstract and extend ArrayList. You then create a inner class in-side the Node which calls `Node.onChildNodeChange()`. Doing so appears to save memory.
* You may also want to wrap the new `children` field in a WeakReference that way our `cache` can be collected if we start running out of memory.

I wont be able to respond for a few days, but I will be happy to clarify what I mean for both of those when i get back.

jhy:
Thanks. I have merged in 38049ce2808cafce3884c25c4613aaaff9f2d0d8. That covers your points on a weak reference (although I used a soft as they're better suited for caches) and moved to the abstract change notifying list for safety.

Thanks again for the code and detail.

Also please do let me know why you're iterating and checking element.is() vs just selecting as it seems that should be faster / less work.

I'll also look to improve the index evaluators so that they don't need to scan, that'll be faster.

jhy:
I removed the auto-boxing on elementSiblingIndex (was returning an Integer but just needed an int) and ops/s went from 450 to 850. 3c81151c7b4dd4739b11f0b6aa5eff81c07fd684

LukeButtersFunnelback:
Cool I looked over the changes they seem really good!

I tried it and the speed up appears to work for master! `43s` down to `1.3s`(it might be faster than that as the machine was doing something else at the time but speed up is clear). I didn't re-do the memory tests, I assume the SoftReference will work.

Did you include a test for the abstract change notifying list?

Why do I user `element.is()`? Basically I have a bunch of things that work on elements themselves to do some verification, however some of these can be skipped as by CSS set by the user so I need to have to check that is `if element.is(userCss) then don't run this`. I guess this is a little odd as i suspect most people have a CSS selector and want to get all matching elements of the selector where I look at all elements and want to know if I should do something different based on some CSS selectors I have.

Here is the sampling of the latest code:

[latest](http://i.imgur.com/xkRZoQI.png)

Thanks for taking the time to work on this to add a change for performance

jhy:
Thanks. I get the use of .is now, makes sense. Always good to get insight into how things are used vs what you imaged when you wrote it :)

I included extra tests to make sure the children elements do get updated correctly, but it's indirect, I'll add some around the abstract list itself as well.

Also I actually moved it back to a WeakReference because those use 32 bytes vs 40 for a SoftReference (there's a timestamp for LRU tracking in there). It actually sped it up marginally as well. Given the use case of many rapid iterations, I found the Weak to be still living long enough between GC cycles to be useful.

LukeButtersFunnelback:
I didn't know about the extra memory used by `SoftReference` that is useful to know.

I tried swapping out Soft and Weak and I wasn't really noticing a consistent enough change to be comfortable claiming one is faster than the other.

PR #864, opened 2017-04-12
Add XmlDeclarationTest.java

leeheekyo:
add src/test/java/org/jsoup/nodes/XmlDeclarationTest.java file.


jhy:
Hi - what is this?

leeheekyo:
It is testcase, but I think it is unnecessary part. So I think you can
ignore that.

2017. 6. 8. 오후 10:34에 "Jonathan Hedley" <notifications@github.com>님이 작성:

jhy:
Ok, thanks. Maybe you could submit a new PR with the specific change and test case?

Issue #872, opened 2017-04-20
Cannot set a request body for HTTP method GET


HexValid:
Hi,

I get this error when I exec this code:

```
Jsoup.connect(URL_API)
                .cookies(this.cookies)
                .ignoreContentType(true)
                .header("Accept", "application/json")
                .header("Content-Type", "application/json; charset=UTF-8")
                .header("Connection", "keep-alive")
                .header("X-CSRF-Token", csrf2)
                .header("X-Requested-With", "XMLHttpRequest")
                .maxBodySize(255)
                .requestBody(json)
                .method(Connection.Method.POST)
                .post();
```

```
Exception in thread "main" java.lang.IllegalArgumentException: Cannot set a request body for HTTP method GET
	at org.jsoup.helper.Validate.isFalse(Validate.java:64)
	at org.jsoup.helper.HttpConnection$Response.execute(HttpConnection.java:639)
	at org.jsoup.helper.HttpConnection$Response.execute(HttpConnection.java:676)
	at org.jsoup.helper.HttpConnection$Response.execute(HttpConnection.java:628)
	at org.jsoup.helper.HttpConnection.execute(HttpConnection.java:260)
	at org.jsoup.helper.HttpConnection.post(HttpConnection.java:255)
```

Why this error, even though I set to `method(Connection.Method.POST)` ?

Thanks

Siedlerchr:
I encountered the same error and I am still wondering

Siedlerchr:
After the req.data.Clear you have to do call `req.requestData(null) `

PR #881, opened 2017-05-10
Set Request body to null when doing a redirecot

Siedlerchr:
Fix for #872 
Set request body to null when doing a redirect

### DIFF
--- a/src/main/java/org/jsoup/helper/HttpConnection.java
+++ b/src/main/java/org/jsoup/helper/HttpConnection.java
@@ -662,6 +662,7 @@ public class HttpConnection implements Connection {
                     if (status != HTTP_TEMP_REDIR) {
                         req.method(Method.GET); // always redirect with a get. any data param from original req are dropped.
                         req.data().clear();
+                        req.requestBody(null);
                     }
 
                     String location = res.header(LOCATION);


jhy:
Thanks, looks good. Do you have an example site I can check on?

Siedlerchr:
Hi,
I need this for the login to sharelatex.com.  In general it should be enough if you do a POST Request to a http website which replies with a 301 response.
The code I used:

```java
    private final String contentType = "application/json; charset=utf-8";
    private final JsonParser parser = new JsonParser();
    private final String userAgent = "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:53.0) Gecko/20100101 Firefox/53.0";
    private Map<String, String> loginCookies = new HashMap<>();
    private String server;
    private String loginUrl;
    private String csrfToken;
    private String projectUrl;

   public String connectToServer(String server, String user, String password) throws IOException {

        this.server = server;
        this.loginUrl = server + "/login";
        Connection.Response crsfResponse;

        crsfResponse = Jsoup.connect(loginUrl).method(Method.GET)
                .execute();

        Document welcomePage = crsfResponse.parse();
        Map<String, String> welcomCookies = crsfResponse.cookies();

        csrfToken = welcomePage.select("input[name=_csrf]").attr("value");

        String json = "{\"_csrf\":" + JSONObject.quote(csrfToken)
                + ",\"email\":" + JSONObject.quote(user) + ",\"password\":" + JSONObject.quote(password) + "}";

        Connection.Response loginResponse = Jsoup.connect(loginUrl)
                .header("Content-Type", contentType)
                .header("Accept", "application/json, text/plain, */*")
                .cookies(welcomCookies)
                .method(Method.POST)
                .requestBody(json)
                .followRedirects(true)
                .ignoreContentType(true)
                .userAgent(userAgent)
                .execute();
```

jhy:
Thanks for that, make sense.

Siedlerchr:
In the linked issue I outlined the problem a bit more in detail:

> The problem is that whenever the request encounters a Location header, it switches the method to GET and wants to execute a 2nd request , but including all previous content from the first request, including the body. Which then leads to the error message.

PR #884, opened 2017-05-15
Java 7 SSLProtocol fix

djechelon:
Hello,
CST Technobank (https://www.csttech.it) is one of the market leaders in Italy in the field of Tax Compliance software for Financial Institution. Our software stack actively uses popular FOSS such as JSoup. We recently had troubles connecting to Italian Revenue Agency (https://telematici.agenziaentrate.gov.it) by means of JSoup at some of our customers running Java 7. We investigated, found a problem with JSoup code and now want to share our findings with the development team.

Inside the HttpConnection class there's `secureProtocol ="SSL"` hardcoded , this means that the TLS version changes according to the jdk. As a result of this with Java 7 projects it's not possible to succesfully connect  to `https` websites that implement a more recent version of it (in our test "https://telematici.agenziaentrate.gov.it/Main/index.jsp" implements TLSv1.2).  There is, however, no problem if the project is ran using Java 8 , which is unsupported by most of our customers.

We have added a method called `secureProtocol` in JSoup API that can be used to set the desired secure protocol, and a Test inside the `HttpConnectionTest` test class  named  setSecureProtocol() that shows it in action. Please check it and share your comments with us.

We would appreciate if JSoup merges this change for the next version. In order to make our software usable before regulatory deadlines, we had to reimplement the above API into a package of ours, thus actually making a fork of JSoup into our code. A really bad workaround.

Credits to
 - Tommaso Formica (tommaso.formica ---at--- csttech.it)
 - Sahaj Quinci (sahaj.quinci ---at--- csttech.it)

### DIFF
--- a/src/main/java/org/jsoup/Connection.java
+++ b/src/main/java/org/jsoup/Connection.java
@@ -1,14 +1,14 @@
 package org.jsoup;
 
-import org.jsoup.nodes.Document;
-import org.jsoup.parser.Parser;
-
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.Proxy;
 import java.net.URL;
 import java.util.Collection;
 import java.util.Map;
+
+import org.jsoup.nodes.Document;
+import org.jsoup.parser.Parser;
 
 /**
  * A Connection provides a convenient interface to fetch content from the web, and parse them into Documents.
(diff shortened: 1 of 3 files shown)


djechelon:
Okay, looks like the tests fail for Java 6.
Actually, it should fail for Java 6 as it doesn't support TLSv1.2.
Source: https://www.ssllabs.com/ssltest/viewClient.html?name=Java&version=6u45&key=25

Maybe we better edit the test to check the Java 6 conditon and not to perform the test in such case

jhy:
Thanks. I don't really understand the problem / proposed solution. Why does this need to be in jsoup? Isn't the solution just placing the updated TLS jar providers from Oracle into the Java secure path for whichever version (I've used 6, 7, 8 with no code changes) is required?

PR #890, opened 2017-05-30
Issue detected when uploading PDF files. 

david-romero:
PDF files were uploaded with a different content type and when downloading them later the browser did not know what content type it is.


jhy:
Thanks, but this diff changes 1K lines. Can you please re-do it so only the changed code is part of the commit?

Also can you give some more details as to what the bug was so we can have a test case?

krystiangorecki:
You can skip showing whitespace changes by adding ?w=1 to diff link, like here: https://github.com/jhy/jsoup/pull/890/commits/231dcbe0bfdd6ed9ce7b2e85e7fc33f4ae2353e7?w=1
Still, unnecessary change in so many lines.

david-romero:
I have created a new pull request to solve problems with code changes.

The new pull request is https://github.com/jhy/jsoup/pull/897

PR #892, opened 2017-06-01
remove Content-Type from get request when redirected

kdaemonv:
I had an issue with an old corporate site (which is out of my control, made probably on ISS with ASP) - it responded 400 for a POST request until I disabled redirects, made POST request, removed the Content-Type header from the connection and requested new location with GET request. This patch helped me to make my code simpler.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,7 +150,7 @@
         <filtering>false</filtering>
         <includes>
           <include>LICENSE</include>
-          <include>README</include>
+          <include>README.md</include>
           <include>CHANGES</include>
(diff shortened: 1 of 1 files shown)


jhy:
Thanks for this. Do you have more detail, what was the Content Type being set to, and by what? Typically it's not a request header. Was it being set by the redirected server response?

Can you send me debug of the request/response/request cycle? Would like to be able to create a test case.

kdaemonv:
Sorry for accidentally closing this - see continuation in PR #895

Issue #883, opened 2017-05-15
Attribute.setKey() is not reliable, key needs to be `final`


kno10:
Assuming we wanted to rename all attributes of a node:
```
for(Attribute at : x.attributes()) {
  at.setKey("old"+ at.getKey());
}
```
will compile, but cause inconsistencies in the `LinkedHashMap<String, Attribute>` od the node, and often show up with the old value afterwards (e.g. when iterating over the LinkedHashMap, and using the original key stored there - because the keys are stored *twice*, once as `String` key in the hash map, once in the `Attribute`).

I think the jsoup API should not allow `setKey` on an attribute, but the key must be considered `final`.

jhy:
Well I'd prefer to just fix the underlying root cause (ie. update the map and the attribute). Otherwise it's not a backwards compatible fix.

kno10:
That will likely cause other problems, such as causing new concurrent modification exception, so it will not be entirely backwards compatible (and in fact, it never really worked reliably!)
The underlying root cause is redundant storage of the key, in the map and the attribute.
```
    Element e = new Element("test");
    e.attr("a", "this is a");
    for(Attribute at : e.attributes()) {
      at.setKey("b");
      at.setValue("b");
    }
    System.out.println("'a' is " + e.attr("a"));
    System.out.println("'b' is " + e.attr("b"));
```
will print
```
'a' is b
'b' is 
```
rather than
```
'a' is 
'b' is b
```

kno10:
#893 is a pull request with a unit test for this bug.

mccxj:
@kno10 trunk code work fine now!

changes at
https://github.com/jhy/jsoup/commit/ea1fb65e9ff8eee82c4e379dc3236d09a5ab02e1

PR #893, opened 2017-06-09
Add unit test (but no fix) for 883

kno10:


kno10:
Confirmed that #883 is fixed now, would you merge this test, too?

jhy:
Thanks, merged

PR #895, opened 2017-06-10
Continuation to PR #892

kdaemonv:
Hi! Content-Type header is set by _HttpConnection.Response#setOutputContentType_ called from _HttpConnection.Response#execute_ when the request method has a body and data was added to the connection (by _Connection#data_).
In my case it was a POST request (it can have a body) and I added data, that's why after _setOutputContentType_ executed I got Content-Type = "application/x-www-form-urlencoded; charset=UTF-8" in the request.
To reproduce - Jsoup.connect("http://some-server-which-returns-400-when-content-type-header-is-present-in-get-request").data("somekey", "somevalue").post();
I created a test case with Jetty as a local server to reproduce the behaviour of the corporate server I encountered the issue with.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -221,6 +221,7 @@
       <groupId>org.mortbay.jetty</groupId>
       <artifactId>jetty</artifactId>
       <version>6.1.26</version>
+      <scope>test</scope>
     </dependency>
 
   </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -217,6 +217,12 @@
       <scope>test</scope>
     </dependency>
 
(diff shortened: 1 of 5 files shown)


jhy:
Digging the Jetty test case, have been thinking of adding that for a bit, and move all the integration tests there,

jhy:
Thanks! I merged the actual code change. I haven't included the Jetty component yet because I want to think more about the structure of it in the project. But definitely like the idea.

kdaemonv:
Ok, thank you for the great library!

PR #897, opened 2017-06-13
Issue detected when uploading PDF files

david-romero:
PDF files were uploaded with a different content type and when
downloading them later the browser did not know what content type it is.

Added test evaluating setOutputContentType method.

If ContentType is PDF bound will be null


jhy:
Thanks. I don't think I understand this change. What's specific about PDFs here? What's the underlying bug / solution?

david-romero:
Hello, 
the underlying bugs is related to the CONTENT-TYPE of the request.

 In this block of code you can see the parameter **bound**, which, in case that's not null, does stuff related to the file's content:

          private static void writePost(final Connection.Request req, final OutputStream outputStream, final String bound) throws IOException {
            final Collection<Connection.KeyVal> data = req.data();
            final BufferedWriter w = new BufferedWriter(new OutputStreamWriter(outputStream, req.postDataCharset()));

            if (bound != null) {
                // boundary will be set if we're in multipart mode
                for (Connection.KeyVal keyVal : data) {
                ....
 
However, when this parameter is bound, in this method:

      private static String setOutputContentType(final Connection.Request req) {
            String bound = null;
            if (req.hasHeader(CONTENT_TYPE)) {
                 // no-op; don't add content type as already set (e.g. for requestBody())
                 // todo - if content type already set, we could add charset or boundary if those aren't included
            }
when the request has a valid *CONTENT_TYPE* and executes the code branch 
`if (req.hasHeader(CONTENT_TYPE))`, 
this leads to the **bound** parameter to be null and fail to write to the request body, resulting in files with no extension (in chrome, firefox, etc) or a blank page with raw data bytes (I think this is related to the *//todo* you have in your code), hence the code I have upload:

       if (req.hasHeader(CONTENT_TYPE)) {
            	//Upload PDF files. If the boundary is not included, the file is uploaded but with an unknown content type.
            	//TODO if it were a known content type, include the boundary
            	if ( APPLICATION_PDF.equals(req.headers().get(CONTENT_TYPE)) ){
            		bound = DataUtil.mimeBoundary();
            	} else {
            		// no-op; don't add content type as already set (e.g. for requestBody())
            		// todo - if content type already set, we could add charset or boundary if those aren't included
            	}
            }

In my issue, I have implemented the case when the Content-type is "application/pdf"; it's left to do other variations of Content-type.

david-romero:
Hi @jhy,

Do you understand my last comment?

I haven't news about this pull request for over a month.

jhy:
There's nothing PDF specific here, so we shouldn't have the PDF check. It looks like we should just move the ifNeedsMulti part out of the else clause.

david-romero:
Move ifNeedsMulti part is a good approach. Maybe, we should be care with the content type. In ifNeedsMulti branch you set content type as "MULTIPART_FORM_DATA".

I'm going to update the PR with this aproach and I hope you can understand the situation.

jhy:
Actually just reviewing this again, and reading the specs. As far as I can tell, the behavior is already correct. The content-type for a multipart upload must be `multipart/form-data`, and the actual content-type for the file is specified on each specific part. Otherwise the server won't know to expect the multiple parts. So with the current implementation, you would do this like

https://github.com/jhy/jsoup/blob/ae7d21635654f1e93e4b732983d633af46f53b73/src/test/java/org/jsoup/integration/ConnectTest.java#L291-L297

Note that the four parameter method is how you specify the part's content-type.

Ref here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

Let me know if I'm missing something!

It probably makes sense to add some example documentation here.

PR #901, opened 2017-06-21
allow template inside thead/tbody/tr. Fixes #807

### LINKED ISSUE #807: Parser error on <template> inside <tr>
I've been experimenting with jsoup as a validator for TensorBoard code and I encountered a bug.

If I have code like this:

```html
          <tr>
            <th></th>
            <th>Name</th>
            <template is="dom-if" if="{{smoothingEnabled}}">
              <th>Smoothed</th>
            </template>   
            <th>Value</th>
            <th>Step</th>
            <th>Time</th>
            <th>Relative</th>
          </tr>
```

I get errors like this:

ERROR: tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.html (offset 1282): Unexpected token [StartTag] when in state [InTable]
ERROR: tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.html (offset 1338): Unexpected token [EndTag] when in state [InTable]
ERROR: tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.html (offset 1338): Unexpected token [EndTag] when in state [InBody]

Please note that those offset numbers point to the `<template>` tags.

Template tag is legal here because https://www.w3.org/TR/html5/tabular-data.html#the-tr-element says content model for `tr` is "Zero or more td, th, and script-supporting elements" and `template` is a script supporting element.

maximermilov:


jhy:
Thanks, merged

jart:
Thank you for the merges today @jhy. You're a saint. As soon as this is one Maven, I'll upgrade all my stuff.

jhy:
(I updated this impl to match the current spec, in #1634)

Issue #363, opened 2013-09-23
Support `file:/...` URLs for local testing


apennebaker:
I'm writing unit tests for an application that uses JSoup, and I would like to write the unit tests in terms of local files mirroring a website rather than the website itself.

As a mitigation, I could write my tests in terms of Files rather than URLs, but this would require substantial refactoring in order to accomplish. Could we please add support for protocols like `file:/...`?

Trace:

```
URL: file:/Users/user/Desktop/src/jsoupcrawler/target/test-classes/google/index.html
java.net.MalformedURLException: Only http & https protocols supported
    at org.jsoup.helper.HttpConnection$Response.execute(HttpConnection.java:417)
    at org.jsoup.helper.HttpConnection$Response.execute(HttpConnection.java:410)
    at org.jsoup.helper.HttpConnection.execute(HttpConnection.java:164)
    at org.jsoup.helper.HttpConnection.get(HttpConnection.java:153)
    at orion.core.data.JSoupCrawler.recursiveCrawl(JSoupCrawler.java:66)
    at orion.core.data.JSoupCrawler.recursiveCrawl(JSoupCrawler.java:19)
    at orion.core.data.JSoupCrawlerTest.testRecursiveCrawl(JSoupCrawlerTest.java:53)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
    at java.lang.reflect.Method.invoke(Method.java:597)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
    at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
    at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
[...]

PR #910, opened 2017-07-04
Support `file:/...` URLs for local testing #363

jjpatel361:
**Usage 1:**
```
String fileURL = "file:///<location_to>.html";
Document doc = Jsoup.allowProtocol(Protocols.FILE, true).load(fileURL);
```
**Usage 2:**
```
Jsoup.allowProtocol(Protocols.FILE, true);
Document doc = Jsoup.load(fileURL);
```
Error:
If accessed without allowProtocol
```java.lang.IllegalAccessError: Enable jsoup to handle file protocol.```

Further protocols can be supported by implementing the ProtocolConnection class
e.g. `FTPConnection` implements `ProtocolConnection`

P.S. This is a crude implementation I'll clean up this PR
Let me know your thoughts on this.

### DIFF
--- a/src/main/java/org/jsoup/Jsoup.java
+++ b/src/main/java/org/jsoup/Jsoup.java
@@ -263,7 +263,8 @@ public class Jsoup {
 	}
 
 	public static Document load(String fileURL) throws Exception {
-		return new FileConnection().load(fileURL);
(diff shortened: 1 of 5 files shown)


jjpatel361:
Would submit with support for `file` and `ftp` protocol

Issue #778, opened 2016-10-26
Method getElementsByIndexLessThan(index) delivers the root element, too


JurMarky:
Hi,
I think this method _getElementsByIndexLessThan(index)_ delivers more than the siblings only.
It delivers the parent of a siblings and the siblings of the parent. so there's confusion with the index-
evaluation.
f.e: [JSoupSiblingTest.zip](https://github.com/jhy/jsoup/files/553360/JSoupSiblingTest.zip)

```
public static void main(String[] args) {
    Document doc = Jsoup.parse(
            "<html><body><div><p>before1</p><p>before2</p><p>XXX</p><p>after1</p><p>after2</p></div></body></html>",
            "");
    Element body = doc.select("body").first();
    Elements elems = body.select("p:matchesOwn(XXX)");
    Element xElem = elems.first();
    Elements beforeX = xElem.parent().getElementsByIndexLessThan(xElem.elementSiblingIndex());

    int i = 0;
    for(Element p : beforeX) {
        p.remove();
    }

    System.out.println(body);


    Document doc2 = Jsoup.parse(
            "<html><body><div><p>before1</p><p>before2</p><p>XXX</p><p>after1</p><p>after2</p></div></body></html>",
            "");
    body = doc2.select("body").first();
    elems = body.select("p:matchesOwn(XXX)");
    xElem = elems.first();
    Elements afterX = xElem.parent().getElementsByIndexGreaterThan(xElem.elementSiblingIndex());

    for(Element p : afterX) {
        p.remove();
    }

    System.out.println(body);
}
```

**Output:**
`LesserThan: <body></body>`
`GreaterThan: <body><div><p>before1</p><p>before2</p><p>XXX</p></div></body>`

**Expected:**
`LesserThan: <body><div><p>XXX</p><p>after1</p><p>after2</p></div></body>`
`GreaterThan: <body><div><p>before1</p><p>before2</p><p>XXX</p></div></body>`

damonwong93:
Just did some digging, and it seems like the root element is actually being added to the beforeX and afterX array lists when you call the getElementsByIndexLessThan() and getElementsByIndexGreaterThan() methods.  This seems to happen because of the matches() method inside the IndexLessThan class, which is called for each element when traversing with NodeTraversor:
```
public boolean matches(Element root, Element element) {
    return element.elementSiblingIndex() < index;
}
```

[...]

PR #914, opened 2017-07-13
Fixed IndexLessThan matches() to ignore root element and added tests #778

damonwong93:
This was in reference to Issue #778, which had problems with the root element being included in a removal of elements with index less than a certain number in a set of siblings.  Modified the matches() validation, so that it doesn't pass when the element being compared to the root is the root.


damonwong93:
Closing this since I created the tests incorrectly.  Will open new pull request later.

PR #921, opened 2017-08-02
Support `file:/...` URLs for local testing #363

jjpatel361:
Added support loading files using `file` protocol. Let me know your thoughts on this.


jjpatel361:
#921

jhy:
Thanks for the PR, and apologies for the late reply. I don't really see the benefit of adding a .load() API.

Why not just use the existing [Jsoup.parse(File file)](https://jsoup.org/apidocs/org/jsoup/Jsoup.html#parse-java.io.File-java.lang.String-) method?

Or, add an option to Connection to allow `file://` protocols, and read it there.

PR #922, opened 2017-08-02
Missing Test Case for location conversion

ileontiuc:
By running mutation testing analysis I discovered that when a jsoup document is being converted into a W3C document the location parsing is never tested. To this end we propose another test case.


jhy:
Thanks!

PR #924, opened 2017-08-02
Missing Test Case for whitespace class name

ileontiuc:
The method that checks if an element has a class is never tested for a class name that ends with a whitespace. The implemented functionality is correct however if the "if (i - start == wantLen && classAttr.regionMatches(true, start, className, 0, wantLen)) {" check is removed, no tests catch this. The missing test case was discovered through mutation testing analysis.


jhy:
Thanks! Merged

PR #925, opened 2017-08-03
Missing Test Case to preserve consumeToIgnoreCase() behavior

ileontiuc:
Since pos is a class attribute, calling consumeToIgnoreCase() on the same queue will return different Strings. If this is the intended behavior than this test preserves that. If not the second assert should be changed to assertEquals("<textarea>one < two ", data); . Nevertheless that would cause the test to fail based on current implementation. This missing behavior preservation was detected by running mutation testing analysis.


jhy:
Thanks, merged. Yes the current behavior is the intended behavior.

Issue #836, opened 2017-02-22
addChildren is quadratic


jart:
A performance improvement was made for `Node#addChildren` a few years ago in #504. But the performance gain only applies when appending nodes without a parent.

Could we make it more like this?

```java
public void addChildren(int index, List<? extends Node> children) {
  for (Node child : children) {
    if (child == null) {
      throw new NullPointerException();
    }
    if (equals(child.parentNode)) {
      throw new IllegalArgumentException();
    }
  }
  ensureChildNodes();
  int oldChildCount = childNodes.size();
  int start = Math.min(0, oldChildCount - children.size());
  for (int i = start; i < oldChildCount; i++) {
    Node child = childNodes.get(i);
    childNodes.add(child);
    child.siblingIndex += children.size();
  }
  for (int i = children.size() - 1; i >= 0; i--) {
    Node in = children.get(i);
    reparentChild(in);
    childNodes.set(start + i, in);
  }
}
```

That would be much faster.

jhy:
That sounds good. Can you let me know your use case - are you moving nodes around? What kind of scale?

(Apologies for the late reply)

mccxj:
@jhy @jart 
patch about  issue #836 addChildren is quadratic
when add many children, the origin implementation will move childNodes array per loop.(method add(index, obj)),
a performance improvement was made using addAll(index, objs).there are three steps:
1. reparent all the children
2. use addAll(index, objs) to insert children
3. reset sibling index
it is very similar with method addChildren(Node... children).

jhy:
Thanks, merged #930

PR #930, opened 2017-08-10
patch about issue #836 addChildren is quadratic

mccxj:
patch about  issue #836 addChildren is quadratic
when add many children, the origin implementation will move childNodes array per loop.(method add(index, obj)),
a performance improvement was made using addAll(index, objs).there are three steps:
1. reparent all the children
2. use addAll(index, objs) to insert children
3. reset sibling index
it is very similar with method addChildren(Node... children).


jhy:
Thanks, merged!

Issue #951, opened 2017-10-09
Method nextElementSibling() returns null after adding an element to a document that was cloned


khdimon:
If I clone a document, and add an element by the method `after()`, and try to get the new element by the method `nextElementSibling()` I get null. In the same time the method `nextSibling()` successfully returns this element.
If I do the same with the original document everything is fine.

Code:

    String html = "<!DOCTYPE html><html lang=\"en\"><head></head><body><div>Initial element</div></body></html>";
    Document original = Jsoup.parse(html);
    Document clone = original.clone();

    Element originalElement = original.body().child(0);
    originalElement.after("<div>New element</div>");
    Element originalNextElementSibling = originalElement.nextElementSibling();
    Element originalNextSibling = (Element) originalElement.nextSibling();
    System.out.println("originalNextElementSibling:\n" + originalNextElementSibling);
    System.out.println("originalNextSibling:\n" + originalNextSibling);
    System.out.println();

    Element cloneElement = clone.body().child(0);
    cloneElement.after("<div>New element</div>");
    Element cloneNextElementSibling = cloneElement.nextElementSibling();
    Element cloneNextSibling = (Element) cloneElement.nextSibling();
    System.out.println("cloneNextElementSibling:\n" + cloneNextElementSibling);
    System.out.println("cloneNextSibling:\n" + cloneNextSibling);

Output:

    originalNextElementSibling:
    <div>
     New element
    </div>
    originalNextSibling:
    <div>
     New element
    </div>

    cloneNextElementSibling:
    null
    cloneNextSibling:
    <div>
    New element
    </div>

jhy:
Thanks for the detail in your report. Fixed!

PR #952, opened 2017-10-09
Bugfix #951 Method nextElementSibling() returns null after adding an element to a document that was cloned 

Joxebus:
This commit fix the ISSUE https://github.com/jhy/jsoup/issues/951

Problem: Once a document is cloned and try to append an after element the document cloned
doesn't recognize the parent and that result in a nextElementSibling equals to null.

Solution: The Element.clone() method calls Element.doClone() with self as parameter to create a new instance.

Changes:
- Element.clone calls to doClone method with self as arg
- New test case added to verify the generated objects are different and with no null values.

"Omar Bautista" <joxebus@gmail.com>

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -1330,6 +1330,6 @@ public class Element extends Node {
 
     @Override
     public Element clone() {
-        return (Element) super.clone();
+        return (Element) super.doClone(this);
     }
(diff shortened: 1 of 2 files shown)


jhy:
Thanks @Joxebus. I also have a fix with c324fabe86c0a59eb635610dc85d8d7a7be25c74. Looks like this change has a problem with cloning the wrong instance.

PR #956, opened 2017-10-22
Pretty print

qxo:


jhy:
Thanks for the PR, and apologies for the late reply. I think this is pretty opinionated, and I don't know that I agree with all the opinions. It would be better to have a discussion (in an issue) with examples of the proposed before and after HTML, and we can review with the larger community how to improve the printer. 

Overall the printer could do with an overhaul and more configuration.

PR #963, opened 2017-11-04
removing last class should remove class attribute

krystiangorecki:
Removes redundant class attribute after removing last class from element.


jhy:
Cool beans, thanks!

PR #969, opened 2017-11-13
Fix #429

### LINKED ISSUE #429: Removing nodes from forms using jsoup
I'm having a problem removing nodes from forms using jsoup v1.7.3. The following code works as expected:

```
Connection.Response response = Jsoup.connect("myURL").execute();

Document doc = response.parse();

//prints HTML including "input[name=alpha]"
System.out.println(doc.toString());

doc.select("input[name=alpha]").first().remove();

//prints HTML excluding "input[name=alpha]"
System.out.println(doc.toString());

FormElement form = (FormElement)doc.select("form").first();

//prints HTML excluding "input[name=alpha]"
System.out.println(form.toString());
```

However, the following code appears to highlight a bug:

```
List<Connection.KeyVal> data = form.formData();

//prints a list including "alpha"
System.out.println(data.toString());
```

I would expect "alpha" to have been removed from the form data, but it hasn't. Is this a bug? Or am I doing something wrong?

[Previously reported on [Stack Overflow](http://stackoverflow.com/questions/24104910/removing-nodes-from-forms-using-jsoup/24110967)]

lukasschlueter:
Calling element.remove() did not remove the element from the form data.

### DIFF
--- a/src/main/java/org/jsoup/nodes/FormElement.java
+++ b/src/main/java/org/jsoup/nodes/FormElement.java
@@ -46,6 +46,12 @@ public class FormElement extends Element {
         return this;
     }
 
+    @Override
+    protected void removeChild(Node out) {
+        super.removeChild(out);
+        elements.remove(out);
+    }
+
     /**
      * Prepare to submit this form. A Connection object is created with the request set up from the form values. You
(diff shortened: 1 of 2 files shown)


jhy:
Thanks! Merged

PR #1000, opened 2018-01-01
Update HttpConnection.java

dty717:
I find a problem in src/main/java/org/jsoup/helper/HttpConnection.java line-769
when contentType="application/javascript"


jhy:
Hi, thanks for the PR. This is the point of the ignore mime type setting - if you want to handle non HTML / XML content, set that to true.

PR #1008, opened 2018-01-19
Handle invalid namespace prefixes when converting to W3C Dom

akbertram:
It is very possible that pages include namespaced tags without
actual namespace definitions. For example, `<ie:menuitem>`

This PR ensures that such malformed elements are gracefully
handled when converting to a w3c object tree.


jhy:
Thanks for the PR, and my apologies for the late reply. I implemented something that has the same effect with a810d2e3615da9a37ad74a7db2ca8bc6945ab9a8

PR #1038, opened 2018-04-07
Accept a custom SSL socket factory

exviva:
This way we can still control the SSL handshake, without disabling certification validation altogether.


jhy:
Thanks, this is great! Will let us deprecate the TLS validation option next round too.

PR #1046, opened 2018-04-15
Parse invalid unicode escapes as Windows-1252 instead [Fixes #1034]

### LINKED ISSUE #1034: Entity parsing of CP-1252 code points
Code points that are illegal characters in unicode (such as `&#0128;`) are parsed by browsers as CP-1252 ([blink](https://chromium.googlesource.com/chromium/blink/+/b69618018614278cda72077611adc093f460dc57/Source/core/html/parser/HTMLEntityParser.cpp#40), [gecko](https://github.com/mozilla/gecko-dev/blob/master/parser/html/nsHtml5NamedCharacters.cpp#L40)). This can lead to certain invalid symbols showing up in browsers but not being parsed in jsoup.

Test case HTML: `&#0128;` [in chromium](https://s.yawk.at/aV1I), [in firefox](https://s.yawk.at/PwKs)

jsoup test case:

```java
import java.util.Arrays;
import org.jsoup.Jsoup;

public final class Test {
        System.out.println(Arrays.toString(Jsoup.parse("&#0128;").text().codePoints().toArray()));
    }
}
```

Prints out `[128]` at the moment, but should be `[8364]` (the € sign).

Unless I missed something this issue looks pretty straight-forward, so I'm willing to write a fix and associated test for this.

yawkat:
See #1034

### DIFF
--- a/src/main/java/org/jsoup/parser/Tokeniser.java
+++ b/src/main/java/org/jsoup/parser/Tokeniser.java
@@ -13,6 +13,17 @@ final class Tokeniser {
     static final char replacementChar = '\uFFFD'; // replaces null character
     private static final char[] notCharRefCharsSorted = new char[]{'\t', '\n', '\r', '\f', ' ', '<', '&'};
 
+    // Some illegal character escapes are parsed by browsers as windows-1252 instead. See issue #1034
+    static final int win1252ExtensionsStart = 0x80;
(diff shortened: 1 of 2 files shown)


jhy:
Thanks! Merged. I found the spec here - https://html.spec.whatwg.org/multipage/parsing.html#numeric-character-reference-end-state

PR #1054, opened 2018-05-02
Add methods to get all the element siblings before or after current element

pipiz:
Currently, element can only get the nearest sibling before or after, but in my actual work, I ofter need to get **all** of it's siblings before or after. So I add this two methods, it's really helpful and convenient.

Hope to merge


pipiz:
Hi, I have modified the code, would you please check if its ok?

jhy:
Thanks! I modified the impl a bit to return Elements instead of List<Element>, and re-used the existing implementation in Elements.

Issue #788, opened 2016-11-16
it I can send  request post multipart/form-data without attachment of the file


Bsplesk:
It I can send  request post multipart/form-data **without attachment of the file** :(

example curl: `curl --form "XXX=YYY" --form "MMM=" --form "SSS=" ...... https://XX.XX.XX.XX/upload`
it is impossible too, most to send jsoap.

[stackoverflow.com](http://stackoverflow.com/questions/40620308/how-to-force-jsoup-to-send-multipart-data?noredirect=1&lq=1)

vcoder4c:
Hi @Bsplesk , did you find the solution for this case? I am trying to do the same thing, but the only solution I can do is custom the header and build the body by myself as multipart/form-data. 

@all: This request have been created for a while, is there any reaction for this?

Bsplesk:
Hi @vcoder4c, It is good that you have opened all your solution.
It will help all who will face this problem.
This problem(and others problem) was solved in 2016, similarly by modifying the source code (unfortunately the source code is not save).
Try to adding your patch to the release (at me it hasn't turned out).
good luck :).

PR #1058, opened 2018-05-07
Force to use multipart/form-data in stead of application/x-www-form-u…

vcoder4c:
Related to the issue #788 , I would like to create this pull request in order to support people who got the problems like the issue's owner and me. It's just a small change, and don't affect to the original function.

### DIFF
--- a/src/main/java/org/jsoup/Connection.java
+++ b/src/main/java/org/jsoup/Connection.java
@@ -495,18 +495,6 @@ public interface Connection {
         Proxy proxy();
 
         /**
-         * Force to use multipart/form-data in stead of application/x-www-form-urlencoded (default)
-         * @return this Request, for chaining
-         */
-        Request forceMultipart();
-
-        /**
-         * Check whether this request is forced to use multipart/form-data in stead of application/x-www-form-urlencoded
-         * @return true if this request is forced to be multipart.
-         */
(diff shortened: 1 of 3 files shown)


jhy:
Thanks for the PR!

I think this would be better implemented by checking to see what content-type the user set. If they set it to multipart, and there's no boundary, add it. That way we're not adding more API surface area, and being more intelligent for the user.

Impl would be around this area:

https://github.com/jhy/jsoup/blob/105f7bdcdbfffc879d0737cf857c02b6823b1b73/src/main/java/org/jsoup/helper/HttpConnection.java#L1004-L1007

Also please make sure you include test cases into ConnectTest.

https://github.com/jhy/jsoup/blob/e26267dd77b437f9e3ea155ae7a52013fde6c2f5/src/test/java/org/jsoup/integration/ConnectTest.java#L35

vcoder4c:
Hi @jhy ,

Thanks for your suggestion. I agree with that, it should be better. And I have already updated. 

1. Revert the old way (open new api to force multipart)
2. Set the boundary when the user set the content type as multipart.

jhy:
Thanks! I squash-merged this in (56ea479e22fdcb360696c5a922d6b45d9445e06d) and changed the constant's visibility for easier access

Issue #1063, opened 2018-05-18
Non-breakable spaces are collapsed into a single space with `.text()`


OndrejSkalicka:
# Summary

Non-breakable spaces (`&nbsp;` / 160 / U+00A0) are considered as non-collapsible in HTML. In d3f0240 this was changed, so now you can either call `TextNode.text()` (and lose nonbreakable spaces) or `TextNode.getWholeText()` (but then you won't get rid of normal spaces).

There was some discussion over the topic in https://github.com/jhy/jsoup/issues/529 , but the solution is not really feasible as you either get "no collapsing" or "collapse everything, even whitespace that should not be collapsed".

# Expected behavior

`TextNode.text()` method to work like it is displayed in browsers. Eg. collapse all whitespace, but do not remove `&nbsp+`.

```java
@Test
void testNbsp() {
    final TextNode textNode = Jsoup.parseBodyFragment("NB&nbsp;&nbsp;&nbsp;SPACES")
            .body().textNodes().get(0);
    // three `&nbsp;` are replaced with three spaces 160/U+00A0 and kept.
    assertThat(textNode.text(), is("NB\u00A0\u00A0\u00A0SPACES"));
}

@Test
void testNormalSpaces() {
    final TextNode textNode = Jsoup.parseBodyFragment("NORMAL   \n\n\t\t SPACES")
            .body().textNodes().get(0);
    // normal spaces are reduced to a single one
    assertThat(textNode.text(), is("NORMAL SPACES"));
}
```

OndrejSkalicka:
Note that you can use this HTML to test out in browser:

```html
<h2>Normal spaced <code style="white-space: pre">foo     bar</code></h2>
<div style="font-family: monospace;">
foo     bar
</div>

<h2>Non-breakable spaced <code style="white-space: pre">foo&amp;nbsp;&amp;nbsp;&amp;nbsp;&amp;nbsp;&amp;nbsp;bar</code></h2>
<div style="font-family: monospace;">
foo&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;bar
</div>

<h2>Mixed spaced <code style="white-space: pre">foo &amp;nbsp;&amp;nbsp;  bar</code></h2>
<div style="font-family: monospace;">
foo &nbsp;&nbsp;  bar
</div>
```

[...]

PR #1071, opened 2018-06-05
fix: Non-breakable spaces are collapsed into a single space with .tex…

dportabella:
fixes this issue:
Non-breakable spaces are collapsed into a single space with `.text()` #1063

### DIFF
--- a/src/main/java/org/jsoup/internal/StringUtil.java
+++ b/src/main/java/org/jsoup/internal/StringUtil.java
@@ -127,8 +127,7 @@ public final class StringUtil {
      * @return true if code point is whitespace, false otherwise
      */
     public static boolean isActuallyWhitespace(int c){
-        return c == ' ' || c == '\t' || c == '\n' || c == '\f' || c == '\r' || c == 160;
-        // 160 is &nbsp; (non-breaking space). Not in the spec but expected.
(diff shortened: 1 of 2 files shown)


jhy:
Hi, thanks for your suggestion. The space normalization in text() is deliberate and is my desired behaviour, as you can see in the texts. People use text() to get strings for indexing, presenting in different contexts, etc. We used to get a lot of feedback / StackOverflows when it wasn't normalized.

How about we add this normalization as a setting in Document.OutputSettings? Make a name like textNbspNormalization.

jhy:
Let me know if you want to make it an option.

dportabella:
Hi, I still think that the default behaviour should be to no collapse non-breakable spaces.
But if you don't agree this this, then making it an option would be the way to go :)

bbulgarelli-yseop:
Hello, could it be possible to plan the making of an option to control the automatic shrink on non breakable spaces ?
On our side, we are handling various user input formats where non breakable spaces are used to indicate specific things, in contrary to the classic char 32 space.
Thanks

bk1991:
So, what is final resolution for this?  text() method is collapsing the whitespaces as well as nbsp, also getWhoteText considering all hardcoded spaced too.
Also cannot find any setting on document outputSetting

Issue #1070, opened 2018-05-30
BOM character is not eaten in Jsoup.parse(String)


amal:
Last versions of Jsoup parse some pages with broken DOM as a result.

Here is an example of such html code:
https://gist.github.com/amal/8d489507c25a329df22a4a0d806aff0b

As a result contents of the `<head>` is not valid, all head tags are in body instead of head:
```html
<html xmlns="http://www.w3.org/1999/xhtml">
 <head></head>
 <body>
  ﻿   
  <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0 maximum-scale=1.0, minimal-ui"> 
  <!-- header tags --> 
  <meta name="guid" content="150305195222805">
  <title>ISIL fighters bulldoze ancient Assyrian palace in Iraq | Iraq News | Al Jazeera</title>
  <meta name="title" content="ISIL fighters bulldoze ancient Assyrian palace in Iraq">
  <meta name="channel" content="">
  <meta name="author" content="">

...
```

krystiangorecki:
I can't reproduce on 1.11.3, 1.11.2, 1.11.1 1.10.2.

```
import java.io.IOException;
import org.jsoup.Jsoup;
import org.jsoup.nodes.Document;

public class JsoupIssue1070 {
	public static void main(final String[] args) throws IOException {
		Document doc = Jsoup.connect("https://gist.githubusercontent.com/amal/8d489507c25a329df22a4a0d806aff0b/raw/38934f389fc300574046080d44338a445e9c9b1e/ancient-assyrian-palace-iraq.html").get();
		System.out.println(doc.select("head>meta").size());
		System.out.println(doc.toString().substring(0, 100));
	}
}
```
and the result always is:
```
30
<!doctype html>
<html xmlns="http://www.w3.org/1999/xhtml">
 <head>
  <meta name="viewport" content=
```

amal:
@krystiangorecki thank you for the test.
Bug seems to be reproducible only with `Jsoup.parse`.
Try this snippet:
```java
import java.io.IOException;
import org.jsoup.Jsoup;
import org.jsoup.nodes.Document;
import org.jsoup.Connection.Response;

public class JsoupIssue1070 {
	public static void main(final String[] args) throws IOException {
[...]

Issue #348, opened 2013-07-29
JSoup incorrectly moves content from the <head> section into <body> for sample URL


franciscrimmins:
If you load the following URL:

```
http://jornutzon.sydneyoperahouse.com/home.htm
```

into:

```
http://try.jsoup.org/
```

then it will move the content from the "head" section into the "body" section. The URL
being parsed validates using the W3C validator:

http://validator.w3.org/check?uri=http%3A%2F%2Fjornutzon.sydneyoperahouse.com%2Fhome.htm&charset=%28detect+automatically%29&doctype=Inline&ss=1&group=0&user-agent=W3C_Validator%2F1.3+http%3A%2F%2Fvalidator.w3.org%2Fservices

We are using JSoup 1.7.2

PR #1073, opened 2018-06-12
[Fix Issue #1070] Jsoup parse string not skipping BOM character correctly

zuozhiw:
Issue #1070 shows a case where Jsoup not parsing an HTML page correctly and putting all the <meta> information in the body instead of head. The problem can be reproducible only if use `Jsoup.parse(response.body())`. Using another method `Jsoup.connect(url).get()` won't reproduce the same problem.

After digging around for a while, it turns out that the HTML page has the BOM character `65279` in its first character. The reason that the second method, `Jsoup.connect(url).get()`, internally uses `DataUtil.parseInputStream()`, which handles the BOM character correctly. However, `Jsoup.parse(string)` constructs a `HTMLTreeBuilder` to parse the document directly, therefore the BOM handling process is not there.

This PR fixes this issue by introducing the same handling process in `TreeBuilder` constructor. Before it passes the input reader to the tokenizer, it reuses the same helper function `DataUtil.detectCharsetFromBom` to detect the BOM character, and skip the BOM character if needed.

This PR also adds a new test case in `DataUtilTest` to test if `Jsoup.parse(string)` can work correctly with BOM. After adding the fix, the problem in issue #1070 can produce the correct result.

There are already several efforts to fix the problem caused by the BOM character:
Issue #348 is the first issue mentioning the problem, and it is fixed in commit https://github.com/jhy/jsoup/commit/3f9f33d88355f22aefc7ea402da09fd1950289ce , this commit only fixes the issue in `parseByteData` (which later becomes `parseInputStream`).

Later there are several commits to refactor the BOM handling code, such as https://github.com/jhy/jsoup/commit/c3cbe1b64e7f66ff9f9b53f1388eb135e6693187 , https://github.com/jhy/jsoup/commit/4eb4f2b2e88a2f9e6c5c1e8d0477060954f24218

The latest issue mentioning the problem is Issue #1003, and the corresponding fix is in https://github.com/jhy/jsoup/commit/0f7e0cc373aced32629f5321c9521f81d8248647 , however, this commit still only fixes the function `parseInputStream`, which is not used by `Jsoup.parse(string)`.

@jhy I'm not sure if `TreeBuilder` class is the best place to put in the fix code. I am willing to devote more efforts to improve it if you could give some more advice. Thanks :)

### DIFF
--- a/src/main/java/org/jsoup/helper/DataUtil.java
+++ b/src/main/java/org/jsoup/helper/DataUtil.java
@@ -238,7 +238,7 @@ public final class DataUtil {
         return StringUtil.releaseBuilder(mime);
     }
 
-    private static BomCharset detectCharsetFromBom(final ByteBuffer byteData) {
+    public static BomCharset detectCharsetFromBom(final ByteBuffer byteData) {
         final Buffer buffer = byteData; // .mark and rewind used to return Buffer, now ByteBuffer, so cast for backward compat
         buffer.mark();
         byte[] bom = new byte[4];
@@ -259,9 +259,9 @@ public final class DataUtil {
         return null;
     }
 
-    private static class BomCharset {
-        private final String charset;
-        private final boolean offset;
+    public static class BomCharset {
+        public final String charset;
+        public final boolean offset;
 
         public BomCharset(String charset, boolean offset) {
             this.charset = charset;
--- a/src/main/java/org/jsoup/parser/TreeBuilder.java
+++ b/src/main/java/org/jsoup/parser/TreeBuilder.java
@@ -1,12 +1,19 @@
 package org.jsoup.parser;
 
+import org.jsoup.UncheckedIOException;
+import org.jsoup.helper.DataUtil;
+import org.jsoup.helper.DataUtil.BomCharset;
 import org.jsoup.helper.Validate;
 import org.jsoup.nodes.Attributes;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
 import org.jsoup.nodes.Node;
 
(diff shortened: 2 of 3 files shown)


panthony:
Any idea when this could be merged ?

panthony:
Using this fix I eventually encountered a nasty issue that made me lost a fair amount of time.

Long story short, I had a piece of code that sometimes where running on a machine that had `US-ASCII` as default charset instead of `UTF-8`.

Because of this, the BOM was badly handled was it relies on `Charset.defaultCharset()`.

This is often bad practice to rely on default Locale/Charset, maybe JSoup could take a look at https://github.com/policeman-tools/forbidden-apis (story behind this project: http://blog.thetaphi.de/2012/07/default-locales-default-charsets-and.html)

panthony:
Not sure which method we are talking about but you cannot convert back the given string to bytes without taking which encoding it was since you cannot rely on `defaultCharset`.

In my case, if `JSoup.parse` were to take an `InputStream` instead of a `String` I could just wrap it inside a `DOMInputStream` (Apache `commons-io`).

Maybe this could just be documented:

- This is the connection layer that handles the BOM implicitly
- The `String` input must be cleared from the BOM markup (only the user can do this since he's the only one to know how it is encoded)
- The `JSoup` could provide an helper that takes a the necessary information (string, charset) to cleanup the string if necessary

jhy:
Closing, as on review I don't plan on taking this. The path through DataUtil which reads the BOM and decodes appropriately is the appropriate way to handle content with BOMs. Just eating it from the start of a String is not appropriate without somehow re-decoding.

Issue #371, opened 2013-11-15
Support HTML redirect (Meta tag refresh) ?


sunng87:
We just got this kind of redirection when parsing html files. Some web sites doesn't use standard HTTP redirection. Instead, they use browser redirection:

``` html
<html>
 <head> 
  <meta http-equiv="Refresh" content="0;URL=http://sports.sina.com.cn/j/2013-11-14/22386885017.shtml" /> 
 </head> 
 <body></body>
</html>
```

We can check the Refresh meta tag: If the content URL doesn't equal to base URL, just treat it as a redirection. This can be done within HTTPConnection. I found a python command line utility [httpie](http://httpie.org) supports this feature. It will be nice to have this feature in Jsoup.

zhuhw:
It does not work for `http://baidu.com` either. The HTML I got is:
```
<html>
 <head>
  <meta http-equiv="refresh" content="0;url=http://www.baidu.com/"> 
 </head>
 <body> 
 </body>
</html>
```
I used `Connection.followRedirects(true)` in version 1.8.3.

jhy:
Closing, no plan to implement.

PR #1086, opened 2018-06-25
Implemented redirect via HTML meta refresh tag support

whyatem:
Resolved issue (#371). As @jhy suggested, this feature is optional and disabled by default. Enable it using `Connection.followMetaRedirects(true)`

### DIFF
--- a/src/main/java/org/jsoup/Connection.java
+++ b/src/main/java/org/jsoup/Connection.java
@@ -120,6 +120,15 @@ public interface Connection {
     Connection followRedirects(boolean followRedirects);
 
     /**
+     * Configures the connection to (not) follow the redirects defined in refresh meta tag of the HTML page.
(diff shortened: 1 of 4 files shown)


jhy:
Hi, thanks for this. Can you add some integration tests, and unit tests for the meta tag -> url parser? I think this only works if the content value is like `0;http://example.com/`. But the spec is `0; url=http://example.com`. So that needs to be more robust.

Also, this will NPE if there is no content attribute and if there is no semi colon in the attribute.

Also can you squash this PR.

Karazuba:
@jhy thanks for pointing everything out, I made some fixes. Not sure how to properly unit test this, due to the fact that the logic is encapsulated, though I did write the integration test.

jhy:
Sorry for holding this one open so long. I'm not going ahead with it. The existing behavior is too well established now, and the instances where that style is used are minimal now IME.

PR #1094, opened 2018-07-15
fix raw text tag content escape issue

qxo:


jhy:
Sorry for the late reply. These are already implemented in:

https://github.com/jhy/jsoup/blob/48d3b2e1600c758f2fa7654a2979ef6cd91073d9/src/main/java/org/jsoup/parser/HtmlTreeBuilderState.java#L530-L544

Confirmed your testcase passes.

Issue #943, opened 2017-09-11
Closed parser/tree builder API


gar1t:
Given the opinionated approach of HtmlParser re strict Html5 structure, which breaks templating schemes (e.g. Polymer/web components - see #600 and #941 and #901, and so on) I need a different tree builder. Unfortunately the tree builder API is closed for extension outside the package. Is there a reason this API is locked down?

jhy:
There's no strong reason, other than the I don't think that the API is very good or one that should be contractual yet. Happy to take suggestions on cleaning it up so we can make it extensible.

jhy:
Closing as not planned. Always happy to hear ideas, please ping me. The core discussed case that I have seen was for `template`, which was implemented in #901 and #1634. 

If there are more cases, it would be good to see if they can be implemented by #2285 Custom Tags.

Issue #600, opened 2015-07-17
Feature Request: Option to Disable Tree Validation in HtmlTreeBuilder


robliao:
HtmlTreeBuilder with HtmlTreeBuilderState performs validation of the tree as it is getting parsed. For example, it restricts the elements available in a table element opting to reparent those into a foster parent when encountered [(Source Link)](https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/parser/HtmlTreeBuilderState.java#L884).

Is it within the bounds of jsoup to provide an option to parse the HTML as is without this reparenting feature?

I should also note that I would like to keep the semantics of HTML Parsing (e.g. data nodes like script element contents are not parsed). This requirement prevents me from using the XmlTreeBuilder.

robliao:
I should also note that I would like to keep the semantics of HTML Parsing (e.g. data nodes like script element contents are not parsed). This requirement prevents me from using the XmlTreeBuilder.

jhy:
What's the use case?

robliao:
I'm working on making the Polymer platform [(Link)](www.polymer-platform.org) workable with the Google Closure Compiler with the Polymer Renamer [(Link)](https://github.com/PolymerLabs/PolymerRenamer).

Polymer's template element provides a way to repeat sections of HTML [(Link)](https://www.polymer-project.org/1.0/docs/devguide/templates.html).

    <template is="dom-repeat" items="{{employees}}">
      <div># <span>{{index}}</span></div>
      <div>First name: <span>{{item.first}}</span></div>
      <div>Last name: <span>{{item.last}}</span></div>
    </template>

This may fall within a table element.

    <table>  
      <template is="dom-repeat" items="{{employees}}">
        <tr>
          <td># <span>{{index}}</span></td>
          <td>First name: <span>{{item.first}}</span></td>
          <td>Last name: <span>{{item.last}}</span></td>
        </tr>
      </template>
    </table>

If this template falls within a table element, Jsoup reparents that to the element containing the table, breaking the template. We'd like to keep this template element where it is.

robliao:
Any update to this? Because we need data note semantics for script and style nodes, there's no available workaround we can use.

jhy:
I don't have an update. I understand the use case, thanks.

venicia:
Hi, we are also needing this feature in our project.  We are using Polymer template tag to iterate data in an array inside the table element.  Any chance the option to disable validation be implemented in November?

[...]

PR #1102, opened 2018-07-24
Extensibility of TreeBuilders

qqilihq:
These changes allow subclassing the TreeBuilder (focus is especially on the `XmlTreeBuilder`, not so much on the `HtmlTreeBuilder`, where I didn't change any of the package-private methods). This includes changing methods of the `Token` class from package-private to public.

Also, I made `outerHtmlHead` and `outerHtmlTail` from `Node` public, to allow building custom toString representations of a Document from custom code.

See related tickets here:

- https://github.com/jhy/jsoup/issues/600
- https://github.com/jhy/jsoup/issues/943

My specific use case: I wanted to process HTML documents for pretty-printing them, but don't want jsoup to modify the existing page structure. Therefore I went with `XmlTreeBuilder` instead of `HtmlTreeBuilder`. The problem: Multiline `<script>` and `<style>` tags would be transformed to a single line with `XmlTreeBuilder`, that's why I needed to create my own customized subclass which would handle those tags the same way as `HtmlTreeBuilder`.

### DIFF
--- a/src/main/java/org/jsoup/nodes/CDataNode.java
+++ b/src/main/java/org/jsoup/nodes/CDataNode.java
@@ -27,14 +27,14 @@ public class CDataNode extends TextNode {
     }
 
     @Override
-    void outerHtmlHead(Appendable accum, int depth, Document.OutputSettings out) throws IOException {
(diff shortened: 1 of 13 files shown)


qqilihq:
Not using this lib any longer. Closing.

PR #1111, opened 2018-08-16
add oraclejdk10 to Travis CI

sullis:


jhy:
Thanks, this is incorporated now

PR #1113, opened 2018-08-20
Add "center" to list of block tags

PumpMagic:
This adds `center` to the list of block tags.

This fixes `Element.text()` for elements that have `<center>`s. For example:

```java
Jsoup.parse("<center>hello</center><center>goodbye</center>").body().text()
```

Before this change, this code evaluates to `hellogoodbye`. With this change, it evaluates to `hello goodbye`.

I tested the change with `mvn test` and it looks like everything is still happy. I also checked the call hierarchy of `Tag.isBlock()` and it looks like this change should be innocuous. Fingers crossed.


jhy:
Thanks, merged. I added a test case - b09e4b7486d8e63b098ff26003b76ed6038f2fb0

PR #1121, opened 2018-09-08
Add 'hasSameValues' method to Elements

HoldYourWaffle:
As per the title.
Allows for easy comparison on matched elements by value instead of identity (which would happen if you use `equals`).

I didn't know how I should document it, so please let me know if that has to be changed.

### DIFF
--- a/src/main/java/org/jsoup/select/Elements.java
+++ b/src/main/java/org/jsoup/select/Elements.java
@@ -1,17 +1,18 @@
 package org.jsoup.select;
 
-import org.jsoup.internal.StringUtil;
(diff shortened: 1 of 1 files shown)


HoldYourWaffle:
Any update on this?

jhy:
Thanks, and sorry for the late reply. I don't know that I immediately see a generally useful feature here / common use case, so I'm currently not inclined to pull it in. Seems like a simple loop suffices here.

Happy to take another look if I've missed something. Would also like to see some tests, and probably call it `haveSameValues` or similar for plural agreement.

PR #1132, opened 2018-10-10
fix problem that same cookie key and different domain cover value.

fonxian:
empty value will cover  the value of not empty string.
e.g.
Set-Cookie: pt_token=abc; path=/; domain=admin.jsoup.org
Set-Cookie: pt_token=; path=/; domain=jsoup.org


jhy:
Thanks, implemented a similar fix with #1476

PR #1136, opened 2018-10-26
redirect when 3xx, enabled, and with Location

abargnesi:
Motivation

The HTTP status codes 300, 301, 303, 307, and 308 allow for client
redirect to the request URL specified in the "Location" header. We
should only redirect when the following conditions are met:
- redirects are enabled (true by default)
- status code is 300, 301, 303, 307, or 308
- a single "Location" header is present in the response

Changes
- Add logic to restrict redirect for the above conditions.
- Updated integration test to check these conditions hold.
- Updated RedirectServlet to allow for multiple "loc" params to be
  configured. This will set multiple "Location" headers in the
  response.

### DIFF
--- a/src/main/java/org/jsoup/helper/HttpConnection.java
+++ b/src/main/java/org/jsoup/helper/HttpConnection.java
@@ -666,8 +666,8 @@ public class HttpConnection implements Connection {
     }
 
     public static class Response extends HttpConnection.Base<Connection.Response> implements Connection.Response {
+        public static final String LOCATION = "Location";
         private static final int MAX_REDIRECTS = 20;
-        private static final String LOCATION = "Location";
         private int statusCode;
         private String statusMessage;
         private ByteBuffer byteData;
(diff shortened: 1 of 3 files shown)


abargnesi:
@Ph3n1x I saw your pull request for #736 and wanted to separate it out. Additionally I have added some tests to cover the conditions.

Is anything missing? How does it look?

mcarleio:
As far as I can see, at least the valid status code 201 is missing: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location

About status code 202, there is some controversy. See here:
* Allowed: https://en.wikipedia.org/wiki/HTTP_location
* Not allowed: https://stackoverflow.com/questions/26199228/is-the-use-of-location-header-in-http-202-response-rfc-compliant

As this change restricts the valid status codes with the Location header (compared to the current behavior) I would also follow the Location header on 202 responses.

Another option would be to make it configureable in some kind of way.

jhy:
Sorry for the late reply, and thanks for your interest. I'm not moving ahead with this - the current behavior is well established now and I haven't seen a requirement to change it.

Issue #1149, opened 2018-11-23
wrong parsing with ParseSettings.preserveCase


anexplore:
jsoup version:1.11.3
when using case sensitive settings,  parse wrong 
```java 
public class TestJsoupParser {

    public static void main(String[] args) {
        Parser parser = Parser.htmlParser();
        parser.settings(ParseSettings.preserveCase); // this line
        String html = "<div class=\"bdsharebuttonbox\">"
                + "<A class=bds_more href=\"http://share.baidu.com/code#\" data-cmd=\"more\">分享到：</A>"
                + "<A title=分享到QQ空间 class=bds_qzone href=\"http://share.baidu.com/code#\" data-cmd=\"qzone\">"
                + "</A><A title=分享到新浪微博 class=bds_tsina href=\"http://share.baidu.com/code#\" data-cmd=\"tsina\"></A>"
                + "<A title=分享到腾讯微博 class=bds_tqq href=\"http://share.baidu.com/code#\" data-cmd=\"tqq\"></A>"
                + "<A title=分享到人人网 class=bds_renren href=\"http://share.baidu.com/code#\" data-cmd=\"renren\"></A>"
                + "<A title=分享到微信 class=bds_weixin href=\"http://share.baidu.com/code#\" data-cmd=\"weixin\"></A>"
                + "</div>\r\n";
        Document doc = Jsoup.parse(html, "", parser);
        System.out.println(doc.html());
    }
    

}
```

the result is:

```
<html>
 <head></head>
 <body>
  <div class="bdsharebuttonbox">
   <A class="bds_more" href="http://share.baidu.com/code#" data-cmd="more">
    分享到：
   </A>
   <A class="bds_more" href="http://share.baidu.com/code#" data-cmd="more">
    <A title="分享到QQ空间" class="bds_qzone" href="http://share.baidu.com/code#" data-cmd="qzone"></A>
    <A title="分享到QQ空间" class="bds_qzone" href="http://share.baidu.com/code#" data-cmd="qzone">
     <A title="分享到新浪微博" class="bds_tsina" href="http://share.baidu.com/code#" data-cmd="tsina"></A>
     <A title="分享到新浪微博" class="bds_tsina" href="http://share.baidu.com/code#" data-cmd="tsina">
      <A title="分享到腾讯微博" class="bds_tqq" href="http://share.baidu.com/code#" data-cmd="tqq"></A>
[...]

PR #1150, opened 2018-11-23
fix case sensitive parse bugs #1149

anexplore:
when call this method param nodeName is normalName() which in lowercase，
when Element tag name is case sensitive, it should use 
```java
node.nodeName().equalsIgnoreCase(nodeName)
```
HtmlTreeBuilder
```java
Element getActiveFormattingElement(String nodeName) {
        for (int pos = formattingElements.size() -1; pos >= 0; pos--) {
            Element next = formattingElements.get(pos);
            if (next == null) // scope marker
                break;
            else if (next.nodeName().equals(nodeName))
                return next;
        }
        return null;
    }
```


anexplore:
this cannot sloved all case sensitive bugs, close it

PR #1151, opened 2018-11-27
fix Parser bugs when use ParseSettings.preserveCase settings #1149

anexplore:
Change:
* add normalName() to Element and Tag
* change HtmlTreeBuilder and HtmlTreeBuilderState to use element's normalName()


jhy:
Hi, thanks for this! I actually didn't see this PR but I did see the bug you filed. I fixed the issue in effectively the same way.

PR #1154, opened 2018-12-05
HTML Parser dropping data while refilling CharacterReader buffer

mboisver:
This is a very rare edge case that we've run into once after extensively using jsoup for several years.  We were seeing that sections of our href query parameters were getting dropped, seemingly at random but with 100% predicability for a given input.  Upon inspection, we saw that an underscore in the middle of our second parameter was coinciding with CharacterReader's readAheadLimit. The reader attempts re-fills its buffer in the middle of reading this parameter.  The problem is that it loses its buffer mark while filling up and drops everything before the underscore in the parameter.

Stack trace for when the mark is reset:
```
	at org.jsoup.parser.CharacterReader.bufferUp(CharacterReader.java:60)
	at org.jsoup.parser.CharacterReader.isEmpty(CharacterReader.java:81)
	at org.jsoup.parser.CharacterReader.matches(CharacterReader.java:353)
	at org.jsoup.parser.Tokeniser.consumeCharacterReference(Tokeniser.java:177)
	at org.jsoup.parser.TokeniserState$38.read(TokeniserState.java:752)
	at org.jsoup.parser.Tokeniser.read(Tokeniser.java:57)
	at org.jsoup.parser.TreeBuilder.runParser(TreeBuilder.java:55)
	at org.jsoup.parser.TreeBuilder.parse(TreeBuilder.java:47)
	at org.jsoup.parser.Parser.parseInput(Parser.java:35)
	at org.jsoup.helper.DataUtil.parseInputStream(DataUtil.java:165)
	at org.jsoup.helper.DataUtil.load(DataUtil.java:54)
	at org.jsoup.Jsoup.parse(Jsoup.java:103)
	at org.jsoup.parser.HtmlParserTest.characterReaderBuffer(HtmlParserTest.java:1217)
```

See attached unit test and HTML sample.  The two links in the source are identical, but the second runs into this case and is impacted by the bug.


jhy:
Thanks for identifyng this and for the test case, it really helped the RCA. I have fixed it with 54a8d301bb486fcc87887d2961da0e82b55dd54e which makes sure that when a mark() is set, a full buffer read happens. That way there's enough data to rewind to if it turns out that what we thought was going to be an HTML entity actually wasn't.

PR #1157, opened 2018-12-13
simplify state sorting status unit test - avoid duplicated code in HtmlTreeBuilderStateTest.java

ny83427:
In test file **HtmlTreeBuilderStateTest** it duplicated tags to search through and there is also a special hint to remind maintainer to synchronize these arrays whenever there is any newly added.

So I made a small improvement to access these arrays via reflection and I tested that the uni test execution time just slow down very little as I executed tests before and after several times.

### DIFF
--- a/src/test/java/org/jsoup/parser/HtmlTreeBuilderStateTest.java
+++ b/src/test/java/org/jsoup/parser/HtmlTreeBuilderStateTest.java
@@ -1,38 +1,27 @@
 package org.jsoup.parser;
 
-import org.jsoup.parser.HtmlTreeBuilderState.Constants;
 import org.junit.Test;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 
 import static org.junit.Assert.assertArrayEquals;
 
 public class HtmlTreeBuilderStateTest {
     @Test
-    public void ensureArraysAreSorted() {
-        String[][] arrays = {
-            Constants.InBodyStartToHead,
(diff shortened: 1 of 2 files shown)


ny83427:
Just found that **HtmlTreeBuilderTest** and **TokeniserStateTest** have similar structure and requirement on newly added string array or char array. It's not a big deal but I think extract these kinds of tests would be beneficial. @jhy What you say?

image

src/test/java/org/jsoup/parser/HtmlTreeBuilderStateTest.java
```diff
@@ -1,38 +1,27 @@
...
-            Constants.InCellTable,
-            Constants.InCellCol,
-        };
+    public void ensureArraysAreSorted() throws ClassNotFoundException, IllegalAccessException {
+        Class constants = Class.forName("org.jsoup.parser.HtmlTreeBuilderState$Constants");
+        Field[] fields = constants.getDeclaredFields();
+        List<String[]> arrays = new ArrayList<>(fields.length);
+        for (Field field : fields) {
+            int modifiers = field.getModifiers();
+            if (Modifier.isStatic(modifiers) && Modifier.isFinal(modifiers) && field.getType() == String[].class) {
```
jhy: My main concern with changing the test to use reflection like this is that the test could, now or in the future, accidentally be skipping fields that need to be tested. E.g. what if down the line we add a constant and for some reason don't mark it final? (I know a bit of a forced example.)

I think if we also add an explicit test that there were _n_ arrays matched it would be more explicit, and can keep a warning in the source to review the test.

Yes, would be good to update the other tests like this, and share a common piece of reflection code.

PR #1162, opened 2018-12-20
Adding a text method with separator to allow splitting of results

1carew1:
I came across the use case of needing to split the result of the text() method but didn't want to extend element and thought this may be a useful addition

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -1042,9 +1042,7 @@ public class Element extends Node {
      * For example, given HTML {@code <p>Hello  <b>there</b> now! </p>}, {@code p.text()} returns {@code "Hello there now!"}
      *
      * @return unencoded, normalized text, or empty string if none.
-     * @see #wholeText() if you don't want the text to be normalized.
-     * @see #ownText()
-     * @see #textNodes()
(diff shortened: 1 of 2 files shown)


avierr:
Thanks for this, jsoup currently returns text separated by a space, while all browsers return the innerText separated by a new line which I find really useful while I parse simple text and don't want to be bothered about the internal node structure.
Ability to specify a custom delimiter is an added bonus as I now can easily distinguish multi line paragraphs.

jhy:
Hi, thanks for this, looks interesting.

The example in the javadoc you included doesn't work. It returns `Hello there now!` Can you check that? Here's some extra test cases that I think would be good to include:

```
        doc = Jsoup.parse("<p>Hello  <b>there</b> now! </p>");
        assertEquals("Hello,,there,,now!", doc.textWithSeparator(",,"));
        assertEquals("Hellotherenow!", doc.textWithSeparator(null));
        assertEquals("Hello\nthere\nnow!", doc.textWithSeparator("\n"));
        assertEquals("Hello$there$now!", doc.textWithSeparator("$"));
```

Also, I think the method name `textSeperatedBy(sep)` might be a bit more active / clearer.

jhy:
Please reopen this if you want to progress it. Thanks!

PR #1164, opened 2018-12-27
add code for consumeCssIdentifier, when attribute class value startswith a dot

anexplore:
```html
<div class=".searchDiv01"></div>
```
for the html code, div's class attr starts with a dot，when use element.cssSelector() method, jsoup will throw a exception caused by:
```java
    private void byClass() {
        String className = tq.consumeCssIdentifier();
        Validate.notEmpty(className); // this line
        evals.add(new Evaluator.Class(className.trim()));
    }
```
so, i add new method to handle this。


jhy:
But, dots are not valid class names. You should use the attribute value selectors instead.

Issue #1116, opened 2018-08-26
feature request: ":has()" and ">"


ogolovanov:
Hello.

It would nice to be able to use "has:()" with ">".
So query like "div:has(> a)" will select all "div" elements that have at lease one direct child "a" element.
Or, for example, add selector like ":hasChild(..)".

Thanks.

jhy:
Thanks, merged in #1214

PR #1175, opened 2019-01-27
Add :hasChild(el) pseudo-selector to support :has(el) selections that search in direct children only, see #1116

haruntuncay:
Added a feature to make :has(el) pseudo selector be able to search only in direct children, as requested in issue #1116.

For example:
`doc.select("div:has(p)") //returns divs that are containing p elements anywhere in their sub-hierarchy.`
New Feature:
`doc.select("div:hasChild(p)") //returns divs that are containing p elements only in their direct children.`

I have added the necessary comment for this new pseudo-selector in Selector.java so the documentation can be updated and added the necessary test case.


haruntuncay:
Hello @jhy ,
Hope you are doing well.
Could you please check out this PR when you have the time.
Thank you in advance.

jhy:
Thanks @haruntuncay, this looks good. We went with #1214 to simplify the selector query syntax.

Issue #1177, opened 2019-02-03
Expand support for java.io.Reader


benmccann:
I see that in https://github.com/jhy/jsoup/issues/463 some support for `Reader` was added: https://jsoup.org/apidocs/org/jsoup/parser/Parser.html#parseInput-java.io.Reader-java.lang.String-

However, I'm not quite sure how to use it. It looks like none of the `Jsoup.parse` methods take a reader. Would it be possible to add a new method taking a `Reader`? I'm trying to use Jsoup inside an interface that only has a `Reader` available

HansBrende:
Just use `Parser.htmlParser().parseInput(reader, baseUri)`.

HansBrende:
You're right though that the functionality to parse a document *fragment* from a reader is missing. That's an open issue I raised some time ago here:  #1143.

benmccann:
Thanks. It might be helpful to document `Parser` if that's the intended method. None of [the examples](https://jsoup.org/cookbook/) refer to it, so it took me awhile to discover

I've sent https://github.com/jhy/jsoup/pull/1178, which adds it to the `Jsoup` class, which seemed to be the intended entry point for all calls from the docs

jhy:
Thanks for the discussion; done!

PR #1178, opened 2019-02-03
Add methods taking a Reader

benmccann:
I'm trying to use Jsoup inside an interface that only has a `Reader` available. It was difficult for me to figure out how to do so without methods on the `Jsoup` class taking a `Reader`

Closes https://github.com/jhy/jsoup/issues/1177


jhy:
Thanks -- I'm not seeing much of a need of these simple delegators - just hit the appropriate parser method directly.

Issue #1186, opened 2019-03-15
Parsing encoding problem with .html() on <style> tag


mino181295:
I have a problem parsing and then setting the _<style>_ tag.
When I do .html() and then set back the string with .html(s) I get the HTML entity &lt; that does not work correctly in CSS.

When I do:
```
        String html = ""
                + "<html>"
                + " <head>"
                + "     <style>.outer > .inner {background-color:white;}</style>"
                + " </head>"
                + " <body>"
                + "     Example"
                + " </body>"
                + "</html>";
        Document document = Jsoup.parse(html);
        Element style = document.selectFirst("style");
        
        String s = style.html();
        //..
        style.html(s);
        
        System.out.println(document.toString());
```
I get:

```
<html>
 <head>
  <style>.outer &gt; .inner {background-color:white;}</style>
 </head>
 <body>
   Example
 </body>
</html>
```

But I'm expecting that the style tag preserve the **>** selector or it won't work correctly.
I'm expecting that the result HTML code is:
```
<html>
 <head>
  <style>.outer > .inner {background-color:white;}</style>
 </head>
 <body>
   Example
 </body>
</html>
```

I would be happy to contribute to an eventual patch if you give me some hints.

mino181295:
@jhy

lnostdal:
Work on a clone in order to avoid mutating the `document` instance. Essentially the same as your code, but in Clojure

```clj
symbolicweb.core> (let [document (Jsoup/parse (str ""
                                         "<html>"
                                         " <head>"
                                         "     <style>.outer > .inner {background-color:white;}</style>"
                                         " </head>"
                                         " <body>"
                                         "     Example"
                                         " </body>"
                                         "</html>"))
                        style (.selectFirst document "style")]
                    (println (.toString (.html style (.html style))))
                    (println "\n###\n")
                    (println (.toString document)))
<style>.outer &gt; .inner {background-color:white;}</style>

###

<html>
 <head> 
  <style>.outer &gt; .inner {background-color:white;}</style> 
 </head> 
 <body>
   Example 
 </body>
</html>
nil
symbolicweb.core> 
```

...when we use .clone, it is fixed:

```clj
symbolicweb.core> (let [document (Jsoup/parse (str ""
                                         "<html>"
                                         " <head>"
                                         "     <style>.outer > .inner {background-color:white;}</style>"
                                         " </head>"
                                         " <body>"
                                         "     Example"
                                         " </body>"
                                         "</html>"))
[...]

PR #1191, opened 2019-03-22
Add Testcase with an Example of CSS Manipulation

eolivelli:
Adding an example usage of CSS manipulation in a testcase.

Relates to #1186


jhy:
Thanks - a couple thoughts:

1. I think it would be better to update the parser / printer to treat style tags as rcdata and not escape that >, vs showing this example workaround
2. The diff contains a lot of unrelated formatting changes, so if we were to proceed with this it would be better to do a clean submission

eolivelli:
@jhy thanks for reviewing my patch


Sorry for the re-formatting.
I will be happy to send a patch with the better approach you are suggesting.
Can you give me some more datailed hint?

PR #1194, opened 2019-03-27
Fix inconsistent javadoc for Element.getElementsMatchingOwnText(String regex)

Anders-E:
Element.getElementsMatchingOwnText(String regex) did not specify it only matches on "own text" making docs a bit confusing when skimming through.


jhy:
Thanks!

Issue #1201, opened 2019-04-06
ShallowClone doesn't clone attributes


iansumm:
When you shallow clone an element you would expect that modifying the attributes of the clone would have no effect on the original element but this is not the case. shallowClone() reuses the same Attributes object from the original element. We should be cloning the Attributes instead. Try the following test.

Element element = new Element("div");
Element clone = element.shallowClone();
clone.addClass("clazz");
assertEquals("", element.className());

PR #1202, opened 2019-04-06
Clone attributes in shallowClone()

iansumm:
Fix for issue #1201 .

When you shallow clone an element you would expect that modifying the attributes of the clone would have no effect on the original element but this is not the case. shallowClone() reuses the same Attributes object from the original element. We should be cloning the Attributes instead.


tamervi:
+1 Issue still exists. Is this issue going to be fixed in the upcoming JSoup release?

jhy:
Thanks! merged

PR #1212, opened 2019-05-13
Disconnect HttpConnection on HttpStatusException and UnsupportedMimeTypeException

sklsop:
Too many open files error could happen here in some circumstances

### DIFF
--- a/src/main/java/org/jsoup/helper/HttpConnection.java
+++ b/src/main/java/org/jsoup/helper/HttpConnection.java
@@ -753,8 +753,10 @@ public class HttpConnection implements Connection {
                     }
                     return execute(req, res);
                 }
-                if ((status < 200 || status >= 400) && !req.ignoreHttpErrors())
-                        throw new HttpStatusException("HTTP error fetching URL", status, req.url().toString());
+                if ((status < 200 || status >= 400) && !req.ignoreHttpErrors()) {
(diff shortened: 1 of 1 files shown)


sklsop:
We had some problems where we parse hundreds of URLs in a row and many of them returned a 404 which caused a too many open files error on some older Samsung Android devices.

jhy:
Hi, thanks for submitting this PR. As far as I can tell, these .disconnects are redundant, because there already is a .disconnect in the IOException catch block (which catches the throws on 759 and 769) on line 799: https://github.com/jhy/jsoup/blob/7de614ffd01161d8718d7458bab9d9690ad2334f/src/main/java/org/jsoup/helper/HttpConnection.java#L796

For sanity I traced execution and verified that the existing disconnect is indeed already called when those exceptions are throw. fd46489cd718ddea7b28f89953265e7ce4ec8372

I have noticed on some Android implementations with okhttp, .disconnect is "advisory" and goes into a different thread to actually reap the connection keepalive. I wonder if that is the root of the problem you are seeing.

Let me know if I'm missing something here or if you have another way to validate this change.

sklsop:
Ups, I didn't realize that.
I'll see if I can find some way do validate it and if so I'll create another pull request. 
Thanks anyway for looking into it!

PR #1214, opened 2019-05-16
Fix :has selector to set the root, allowing relative selectors

### LINKED ISSUE #1116: feature request: ":has()" and ">"

lumimies:
Fixes #1116


jhy:
Thanks! Great simple fix, and good to include the test case. Merged.

jhy:
Merged in 0fe6d51fd8b026e6a860376b228533da66654fb4

ogolovanov:
Hi!

Does this really works?

I have installed jsoup 1.2.2 from git.

And get errors like:
Could not parse query 'div:hasChild(h2)': unexpected token at ':hasChild(h2)'

PR #1217, opened 2019-05-21
Jdk11 / Jdk12 JavaDoc generation fix

offa:
Fix for the JavaDoc generation failure on JDK11 and JDK12 (OpenJDK bug [**JDK-8212233**](https://bugs.openjdk.java.net/browse/JDK-8212233)).

In addition, update to _Maven JavaDoc Plugin v3.1.0_ and the new `doclint` option, introduced with v3.0.0.


jhy:
Thanks! Merged

Issue #1218, opened 2019-05-21
Websites with large amounts of data fail to parse.


Derek-Baum:
Currently using Jsoup on some large websites, and it throws the Mark Invalid Exception which means the bufref is negative?

I tried using both Jsoup.connect(url).get()
and Jsoup.connect(url).execute().parse()
Both cause the same exception.
```Exception in thread "main" java.io.IOException: Mark invalid
	at org.jsoup.parser.CharacterReader.rewindToMark(CharacterReader.java:132)
	at org.jsoup.parser.Tokeniser.consumeCharacterReference(Tokeniser.java:182)
	at org.jsoup.parser.TokeniserState.readCharRef(TokeniserState.java:1698)
	at org.jsoup.parser.TokeniserState.access$100(TokeniserState.java:8)
	at org.jsoup.parser.TokeniserState$2.read(TokeniserState.java:36)
	at org.jsoup.parser.Tokeniser.read(Tokeniser.java:57)
	at org.jsoup.parser.TreeBuilder.runParser(TreeBuilder.java:55)
	at org.jsoup.parser.TreeBuilder.parse(TreeBuilder.java:47)
	at org.jsoup.parser.Parser.parseInput(Parser.java:35)
	at org.jsoup.helper.DataUtil.parseInputStream(DataUtil.java:169)
	at org.jsoup.helper.HttpConnection$Response.parse(HttpConnection.java:835)
	at org.jsoup.helper.HttpConnection.get(HttpConnection.java:285)```

if anybody would like to reproduce, here are some urls which it fails to parse:
https://www.spec.org/cpu2006/results/res2014q4/
https://www.spec.org/cpu2006/results/res2012q3/
https://www.spec.org/cpu2006/results/res2014q1/
https://www.spec.org/cpu2006/results/res2014q3/
https://www.spec.org/cpu2006/results/res2011q2/
https://www.spec.org/cpu2006/results/res2010q3/
https://www.spec.org/cpu2006/results/res2017q2/
https://www.spec.org/cpu2006/results/res2016q3/
https://www.spec.org/cpu2006/results/res2015q4/
https://www.spec.org/cpu2006/results/res2007q4/
https://www.spec.org/cpu2006/results/res2009q4/
https://www.spec.org/cpu2006/results/res2012q2/
https://www.spec.org/cpu2006/results/res2014q2/
https://www.spec.org/cpu2006/results/res2012q4/
https://www.spec.org/cpu2006/results/res2011q1/


Thanks

[...]

PR #1225, opened 2019-06-01
Proposed fix for #1218 ("Mark invalid" error message)

csaboka:
The current parsing code gives up parsing in certain cases instead of reading more data from the underlying reader. This can cause valid markup to be misparsed. Additionally, marks can be invalidated when more data is pulled into the buffer, so marking and rewinding works unreliably.

My attempt of fixing is required some refactoring, so the various consumeXxx() methods keep going after reaching the end of the buffer, maintaining a temporary StringBuilder to hold the data seen so far. To fix marking and rewinding, all the data past the marked position is kept in the buffer when reading more data. Now the client needs to explicitly release the mark to stop it from being kept in the buffer.

Unit tests are still all green, but this definitely needs a review from someone more familiar with the project.

### DIFF
--- a/src/main/java/org/jsoup/parser/CharacterReader.java
+++ b/src/main/java/org/jsoup/parser/CharacterReader.java
@@ -16,12 +16,10 @@ public final class CharacterReader {
     static final char EOF = (char) -1;
     private static final int maxStringCacheLen = 12;
     static final int maxBufferLen = 1024 * 32; // visible for testing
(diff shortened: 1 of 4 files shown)


jhy:
@csaboka thank you again for your work and help on this! I ended up taking a slightly different approach (as you've seen) in an attempt to keep the charbuf reads minimized, but your code very much helped my approach.

Thanks,
Jonathan

PR #1226, opened 2019-06-01
Warning fixes

TheRamenChef:
This pull request makes several changes to fix warnings and better comply with Java best practices. This includes:

  - Adding `SerialVersionUID` fields
  - Adding type parameters where missing
  - Adding `@Override` annotations

Among several other things.


jhy:
Sorry for the late reply -- this conflicts now so won't apply, but generally I think this is too big of a diff to review, and I don't see much appreciable value from adding all of these changes.

Issue #1228, opened 2019-06-04
Create a setter method to sets Connection read timeout


kobikis:
In the following example the Connection timeout is 7sec and the url latency is 5sec but i'm getting java.net.SocketTimeoutException: Read timed out

`
@RunWith(SpringRunner.class)
public class FetcherConcurrentClientApplicationTests {

   @Test
   public void simpleTest() {
      Document doc = null;
      long startTime = System.nanoTime();
      try {
         doc = Jsoup.connect("http://httpstat.us/200?sleep=5000").timeout(7000).get();
      } catch (IOException e) {
         e.printStackTrace();
      }
      long processingTime = System.nanoTime() - startTime;
      System.out.println("Processing time: " + TimeUnit.NANOSECONDS.toMillis(processingTime));

      assertNotNull(doc);
   }
}
`

It's because that when JSoup create a connection it configure the Connection readTimeout to half of the request timeout.

`
private static HttpURLConnection createConnection(Connection.Request req) throws IOException {
            final HttpURLConnection conn = (HttpURLConnection) (
                req.proxy() == null ?
                req.url().openConnection() :
                req.url().openConnection(req.proxy())
            );

            conn.setRequestMethod(req.method().name());
            conn.setInstanceFollowRedirects(false); // don't rely on native redirection support
            conn.setConnectTimeout(req.timeout());
            conn.setReadTimeout(req.timeout() / 2);
`

jhy:
See #1883. I've added a note in the javadoc to clarify the behaviour.

I don't intend to add a socket read timeout setting specifically, as there is no setting for that when using the HttpClient which is now our default.

PR #1229, opened 2019-06-04
Create a setter method to sets Connection read timeout #1228

kobikis:


jhy:
Thanks -- I don't currently plan to merge this but I do want to revamp how the timer is run. And then we might provide explicit settings for both connect and read. But I would prefer ultimately to only have the total timeout, and make connect the larger portion of that. And when using a shorter read timeout, recover from a socket timeout if there is still time remaining in the total timer. Not sure how possible that is, both in the HttpURLConnection and the HttpClient impl, but would be good to find.

PR #1242, opened 2019-07-16
Proposed fix for #1218, take 2

csaboka:
I didn't understand how the overall parsing works when opening #1225 , so I tried to solve all the issues in CharacterReader. This PR has a less intrusive approach that alters the behavior of bufferUp() instead of adding extra buffering to various methods.

Also contains added tests for bugs I've found while investigating the code and proposed fixes for them.

### DIFF
--- a/src/main/java/org/jsoup/parser/CharacterReader.java
+++ b/src/main/java/org/jsoup/parser/CharacterReader.java
@@ -55,9 +55,9 @@ public final class CharacterReader {
         try {
             final long skipped = reader.skip(pos);
             reader.mark(maxBufferLen);
-            final int read = reader.read(charBuf);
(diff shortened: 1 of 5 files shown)


HansBrende:
@csaboka I'm wondering, do your changes also fix https://github.com/jhy/jsoup/issues/1251 ?

csaboka:
> @csaboka I'm wondering, do your changes also fix #1251 ?

Sorry, but apparently not. When I add your repro code as a unit test, it still fails with "No buffer left to unconsume". I don't have time right now to investigate it further, but maybe during the weekend.

csaboka:
Closing this since the original issue is solved now. See #1303 for the minor fixes that were introduced by this PR but could be still useful now.

Issue #1227, opened 2019-06-03
Incorrect Farsi parsing to plain text


MarcFletcher:
Both are expected to produce the same output, although the first one incorrectly strips the 'half-space' `(\u200C)`:

    System.out.println("Incorrect: " + Jsoup.parse("نیمه\u200Cشب").text());
    System.out.println("Correct: " + "نیمه\u200Cشب");

Output:

![Sample Output](https://i.imgur.com/7hfkrIe.png)

In Farsi this is quite a considerable change of meaning.

jhy:
This comes from #978, where we strip invisibles from text, so matching is easier (removing soft hyphens in that case).

This character is not a half-space though, it's the zero-width-non-joiner.

https://unicodelookup.com/#0x200c/1
https://en.wikipedia.org/wiki/Zero-width_non-joiner

Probably need to review what gets stripped in the invisibles method. &shy is OK to strip as it is more of a render time thing (width of browser) vs semantic as in this example.

jhy:
Thanks, fixed with https://github.com/jhy/jsoup/commit/91ca25b341bc5ad1c364b8e7389287c45ca9df2c

PR #1248, opened 2019-08-14
Do not consider zero width non-join character as invisible

diepjy:
This pull request should fix the issue where the zero width non join character is getting removed by `StringUtil.isInvisibleChar` when it is significant semantically for some languages for example Farsi.

This is a fix for issue: https://github.com/jhy/jsoup/issues/1227

Have updated `StringUtil.isInvisibleChar` to ensure the ZWNJ doesn't get stripped out and added/updated unit tests accordingly.


jhy:
Thanks! Sorry I overlooked this and implemented essentially the same with https://github.com/jhy/jsoup/commit/91ca25b341bc5ad1c364b8e7389287c45ca9df2c

PR #1253, opened 2019-08-31
Report the correct error position in some malformed constructs.

### LINKED ISSUE #1251: Bad html causes ArrayIndexOutOfBoundsException (v1.11.3), UncheckedIOException (v1.12.1)
```java
public void test() {
    byte[] data = new URL("https://github.com/apache/any23/files/3551705/weirdjsoupmin.txt").openStream().readAllBytes();
    Parser.parseFragment(new String(data, StandardCharsets.ISO_8859_1), null, "", ParseErrorList.tracking(1));
}
```

Output (jsoup v1.11.3):
> Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: -1
	at org.jsoup.parser.CharacterReader.consume(CharacterReader.java:99)
	at org.jsoup.parser.TokeniserState$34.read(TokeniserState.java:552)
	at org.jsoup.parser.Tokeniser.read(Tokeniser.java:57)
	at org.jsoup.parser.TreeBuilder.runParser(TreeBuilder.java:51)
	at org.jsoup.parser.HtmlTreeBuilder.parseFragment(HtmlTreeBuilder.java:126)
	at org.jsoup.parser.Parser.parseFragment(Parser.java:138)

Output (jsoup v1.12.1):
> Exception in thread "main" org.jsoup.UncheckedIOException: java.io.IOException: No buffer left to unconsume
	at org.jsoup.parser.CharacterReader.unconsume(CharacterReader.java:111)
	at org.jsoup.parser.TokeniserState$41.read(TokeniserState.java:877)
	at org.jsoup.parser.Tokeniser.read(Tokeniser.java:57)
	at org.jsoup.parser.TreeBuilder.runParser(TreeBuilder.java:55)
	at org.jsoup.parser.HtmlTreeBuilder.parseFragment(HtmlTreeBuilder.java:126)
	at org.jsoup.parser.Parser.parseFragment(Parser.java:140)
Caused by: java.io.IOException: No buffer left to unconsume
	... 7 more

This exception originally occurred when parsing the webpage https://www.streetpadel.com/overgrip-head-pro-grip-dz-negro-p-17233.html but I whittled the html down to a minimal example that reproduces the issue, here: https://github.com/apache/any23/files/3551705/weirdjsoupmin.txt. 

Cf. [ANY23-441](https://github.com/apache/any23/pull/139)

csaboka:
If the consumed character is unexpected, we need to unconsume it first,
then report the error, so the correct character is reported. In rare
cases, the extra buffering up caused by reporting the invalid character
could cause unconsume() to fail entirely.

Fixes #1251.


src/test/java/org/jsoup/parser/TokeniserStateTest.java
```diff
@@ -1,209 +1,252 @@
-package org.jsoup.parser;
```
csaboka: I'm not sure why the diff shows the entire file is replaced. I've only added new tests, starting from line 210.

csaboka:
The build failure doesn't seem to have anything to do with my changes:
`No JDK version is compatible with amazon-corretto@1.8.212-04.2`

src/main/java/org/jsoup/parser/TokeniserState.java
```diff
@@ -565,8 +565,8 @@ void read(Tokeniser t, CharacterReader r) {
                     t.transition(SelfClosingStartTag);
                     break;
                 case '<': // NOTE: out of spec, but clear (spec has this as a part of the attribute name)
-                    t.error(this);
                     r.unconsume();
+                    t.error(this);
```
HansBrende: Should this also be changed in the `nullChar` case 6 lines below this line?

HansBrende:
@csaboka I've tested your fix and it seems to work. Only question I have is that one comment I added to the diff.

src/main/java/org/jsoup/parser/TokeniserState.java
csaboka: You're right. I'll fix this.

csaboka:
The Travis build is still failing for reasons unrelated to the Java source, but as far as I can tell the code is actually working.

HansBrende:
@csaboka I investigated why that whole test file shows up as replaced in the diff. Apparently you inadvertently replaced all the newlines with carriage return + line feed (the default on windows). Maybe edit that file to use `\n` instead of `\r\n` as the line separator?

csaboka:
@HansBrende I don't get it. I am on Windows, and looks like my IDE didn't pick up the intended core.autocrlf setting, but now I've fixed it and the diff still reports all lines changed. When I force core.autocrlf to false and get the file again from my index, it has \n line endings everywhere, so the pushed version should have \n endings as well. I'm out of ideas.

HansBrende:
Nope, it still shows `\r\n` everywhere. You can see this by copy-and-pasting the raw text to a website that shows hidden characters, e.g. https://www.soscisurvey.de/tools/view-chars.php

src/test/java/org/jsoup/parser/TokeniserStateTest.java
```diff
@@ -1,209 +1,252 @@
...
-    }
-
-    @Test public void handlesLessInTagThanAsNewTag() {
-        // out of spec, but clear author intent
-        String html = "<p\n<p<div id=one <span>Two";
-        Document doc = Jsoup.parse(html);
-        assertEquals("<p></p><p></p><div id=\"one\"><span>Two</span></div>", TextUtil.stripNewlines(doc.body().html()));
-    }
-}
+package org.jsoup.parser;
```
HansBrende:
```suggestion
package org.jsoup.parser;
```

src/main/java/org/jsoup/parser/TokeniserState.java
```diff
@@ -894,8 +894,8 @@ void read(Tokeniser t, CharacterReader r) {
                     t.transition(Data);
                     break;
                 default:
-                    t.error(this);
                     r.unconsume();
+                    t.error(this);
```
HansBrende:
```suggestion
                    t.error(this);
```

HansBrende:
@csaboka This question might help you https://stackoverflow.com/questions/27810758/how-to-replace-crlf-with-lf-in-a-single-file

csaboka:
@HansBrende : I appreciate that you're trying to help, but please believe me that I already know how to use a hexeditor and verify which kind of line endings I have. When I check out the error-reporting-fix branch on my local repo, I'm getting Unix-style endings (\n, 0x0A) even when I have the "core.autocrlf" setting turned on in Git, which is strange. In any case, I can't convert the file to LF endings when it already has all LF endings.

jhy:
Thanks @csaboka and @HansBrende! Have merged this in c2b1fe73129e9c23d93e5393ce8007995c1630dc

I couldn't figure what was going on with that diff either, so I ended up just C&Ping in the new tests. @csaboka if you don't mind, could you please double check that I pulled everything in fully?

csaboka:
@jhy : The commit looks good to me, it contains all my intended changes and all the new tests.

PR #1254, opened 2019-09-10
CI builds for current Java versions

offa:
CI builds for Java 12 (latest) and Java 13 (will be released within a week).

The amazon corretto version isn't available anymore so I had to update that too.


offa:
@jhy How about using docker container for the ci build?

jhy:
Thanks, merged. I updated the config spec so it would be less sensitive to minor version revs. 

Docker is a good idea, but I do like that it's simple to add new JDKs here.

PascalSchumacher:
Java 9 and 12 are non-long-term support versions which are no longer supported. Imho it is not necessary to test with these versions.

By the way: Thank you very much for jsoup!

PR #1256, opened 2019-09-19
Seamless JSON support for Jsoup (v2.0 "less code")

magicprinc:
1. Powerful JSON parser for Jsoup (can consume every text, doesn't throw exceptions).
2. 99% code coverage.
3. 3-10 times faster than XmlTreeBuilder.
4. Very convenient way to query large json documents with Jsoup selectors (I use it with 20MB large json).

I haven't added JSON content type autodetection: for backward compatibility, user must set it manually.

Update:
Very small changes in Jsoup classes (private to package visibility for 2 methods and jsonParser method moved to Parser class), but less code in JsonTreeBuilder.


magicprinc:
"build failed" only for:
JDK="amazon-corretto@1.8.212-04.2"

because:

Computing best match for required JDK version: amazon-corretto@1.8.212-04.2
Selected JDK: 
No JDK version is compatible with amazon-corretto@1.8.212-04.2. Available JDKs are:
...
The command "bash .install-jdk-travis.sh" failed and exited with 2 during .
Your build has been stopped.

jhy:
Hi @magicprinc, this looks cool. However I feel it falls outside of the scope of the core jsoup project. And, we can't pull in copyrighted code or that has a non-MIT license. It might be an idea to pull this out into a seperate project (which would have to be under the Apache 2 license if you use the same code)

magicprinc:
I have finally released it as a separate project. 
See #1679

PR #1258, opened 2019-09-25
Adjustable case-sensitivity for whitelist

cg-swiftlox:
The changes in the whitelist class allow to adjust the case-sensitivity for checking tags, attributes and protocols.


jhy:
Thanks, but I'm not seeing the value of this as a feature (and it's not described). Seems better just to choose to preserve / not-preserve case at the output-level, vs stripping tags based on case.

Apologies for the late review.

PR #1261, opened 2019-10-01
Added additional overload for connection

sanyarnd:


jhy:
Thanks, but seems a bit superfluous IMV

PR #1263, opened 2019-10-09
Added a test case for connecting through a HTTP proxy.

theaveragedude:
The proxy used in the test was obtained from a free proxy list found on the internet.
There's no guarantee that the proxy server would respond next time the test is run.
It would be better to replace the free proxy with a persistent one.


jhy:
Thanks for the suggestion. I've been working to remove any external hits in the tests. Certainly I'm not going to add dependencies on these. I'd suggest refactoring and using the local Jetty server as a proxy.

jason-327:
请教下用了代理一直报：
java.net.ConnectException: Connection refused: connect
image

是什么情况

Issue #1106, opened 2018-08-07
Comments may contain invalid characters that should have been converted to entities


hannibal218bc:
When setting an output charset, Jsoup does not automatically convert characters not supported by the target charset to entities. The resulting document thus contains illegal characters.

Maybe `Comment.getData()` should also use `Entities.escape() like `TextNode` ?

See this live example: http://try.jsoup.org/~e0Nc-LMkpYKPAHhbnuGUICq-4g4
The "ascii" option is activated, so the target document should be valid ASCII. However, it still contains äöü Umlauts.

hannibal218bc:
I see that this issue has not yet been addressed. Is the suggested fix not appropriate?

jhy:
This is the same issue at #1271. When you parse, comments don't get decoded as they are data, so it doesn't make sense to encode them.

hannibal218bc:
OK, agreed by the same reasoning as https://github.com/jhy/jsoup/issues/1271#issuecomment-748521612

PR #1272, opened 2019-10-19
Comment must use Entities.escape()

hannibal218bc:
Suggested fix for #1106


jhy:
Closing as per bug; comments aren't decoded on parse and so shouldn't be encoded

Issue #1279, opened 2019-11-19
Starts with selector not working


andrfgs:
I am using JSoup 1.11.3 and everything works flawlessly except for the **starts with** selector.

The following selector `a[href ^= ' mailto']` does not yield any results and throws exception:
> Exception in thread "main" java.lang.IndexOutOfBoundsException: Index: 0, Size: 0

This is my code:
`String str = d.select("a[href ^= ' mailto']").get(0).html();`

Running the selector on Opera (Webkit), yields the correct result (by typing `$("a[href ^= ' mailto']")` in the console)

This is the page I am parsing: https://pastebin.com/si1Zy2cz

jhy:
Thanks, confirmed and merged

andrfgs:
Thank you a lot for all your work maintaining this library!

PR #1280, opened 2019-11-22
fix for bug #1279

michelpetrovic:
The general idea is that the value of the `startswith`-selector may contain spaces if it is inside quotes, in which case we don't trim the value.


jhy:
Thanks! Merged. I added a fix for the same issue with "Ends With Space" as well.

PR #1284, opened 2019-12-06
Node traversor

hsebs:


hsebs:
It is some mistakes. This branch is not valid.

PR #1291, opened 2020-01-01
Added new method childrenCount complementary to child(int)

literakl:
There is no easy way in the Element to find out a number of children available via the method child(int). The method childNodeSize() acts on list of all children while child(int) acts on the filtered list of the children. This trivial patch adds new method childrenCount() that is complementary to child(int) as it works on the same filtered list. A unit test is provided.

My observation: https://stackoverflow.com/questions/59547574/jsoup-how-to-find-out-elements-size

Thank you for this wonderful library and for consideration of merging this proposal.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -265,6 +265,18 @@ public class Element extends Node {
     }
 
     /**
+     * Get the number of child nodes that are elements.
+     * <p>
+     * This method works on the same filtered list like {@link #child(int)}.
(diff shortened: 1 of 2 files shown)


literakl:
It seems there is some configuration issue on github that the build fails:

The command "bash .install-jdk-travis.sh" failed and exited with 2 during .

jhy:
Thanks! Merged with 894d3ec9c71a324c007a4790604e6ca44e1d83c7

I renamed it to be `childrenSize` for consistency with the other sizes of lists.

Issue #1300, opened 2020-01-16
UnknownHostException for cyrillic url


paveloso:
Document doc = Jsoup.connect("http://en.wikipedia.org/").get();

if you pass the string to the connect() method with the cyrillic url in it (ex. "http://что-угодно.рф"), the jsoup throws the UnknownHostException with the passed link encoded to another encoding.

Any work arounds?

paveloso:
found an issue. The issue happens because of IDN.

made a fix here (#1301 ) that encodes input URL into Punycode with additional parameter.

PR #1301, opened 2020-01-17
allows to use IDN to connect to and get Document

paveloso:
added new param in Jsoup.connect method in order to connect to IDN. This translates the URL to Punycode and at the end return the Document with data.


paveloso:
the issue for this is #1300

jhy:
Thanks for this, @paveloso. This is failing because jsoup asserts that it runs on Android API level 8, and per the [Android IDN](https://developer.android.com/reference/java/net/IDN) doc, the min API version is 9.

Now the [Android distributions counts](https://developer.android.com/about/dashboards) don't even start before 10, so it's probably safe to upgrade the base requirement to at least API v 10.

src/main/java/org/jsoup/helper/HttpConnection.java
```diff
@@ -20,14 +20,7 @@
...
 import java.io.UnsupportedEncodingException;
-import java.net.HttpURLConnection;
-import java.net.InetSocketAddress;
-import java.net.MalformedURLException;
-import java.net.Proxy;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.net.URLEncoder;
+import java.net.*;
```
jhy: It's best to keep all imports explicit.

src/main/java/org/jsoup/Jsoup.java
```diff
@@ -73,6 +73,22 @@ public static Connection connect(String url) {
...
+     * <ul>
+     *  <li><code>Document doc = Jsoup.connect("http://example.com").userAgent("Mozilla").data("name", "jsoup").get();</code></li>
+     *  <li><code>Document doc = Jsoup.connect("http://example.com").cookie("auth", "token").post();</code></li>
+     * </ul>
+     * @param url URL to connect to. The protocol must be {@code http} or {@code https}.
+     * @param toPunycode a flag if you need to encode the URL into Punycode in order to connect to IDN (Internationalized Domain Names).
+     * @return the connection. You can add data, cookies, and headers; set the user-agent, referrer, method; and then execute.
+     */
+    public static Connection connect(String url, Boolean toPunycode) {
+        return HttpConnection.connect(url, toPunycode);
```
jhy: We shouldn't have this method - just punycode all URLs as they come in. Otherwise the user has to know that it's an IDN in the URL for every call, and remember this, and all the link following and redirect code will fail.

src/test/java/org/jsoup/helper/HttpConnectionTest.java
```diff
@@ -160,6 +161,13 @@
         assertEquals("http://example.com", con.request().url().toExternalForm());
     }
 
+    @Test public void connectWithIdnUrl() throws MalformedURLException {
+        URL url = new URL("https://тест.рф");
+        Connection con = HttpConnection.connect(url, true);
```
jhy: Need to make sure we have test for redirects to IDNs, and for finding IDN in HTML and following.

src/main/java/org/jsoup/helper/HttpConnection.java
paveloso: sure, it's just IDE combined those

src/test/java/org/jsoup/helper/HttpConnectionTest.java
paveloso: Can't write tests. Have lack of knowledge.

src/main/java/org/jsoup/Jsoup.java
paveloso: removed

jhy:
Thanks. I implemented with similar checkin - 21b01b258952fd6aeb55d061f98321e5b9fe93e0

PR #1303, opened 2020-01-21
Fix edge cases with delimiters being near a buffer boundary.

csaboka:
These changes are fixes for issues I've found while working on #1242. The original issue I've opened #1242 for is solved now, but I think it's worthwhile to merge these other minor fixes.


jhy:
Looks great, thanks!

Issue #1305, opened 2020-01-22
prettyPrint and outline surround inline-tags with white space (where they shouldn't)


EdyHinzen:
Using this code with jsoup 1.11.3:

```
String html = "<div><span>1:15</span>&ndash;<span>2:15</span>&nbsp;p.m.</div>";
org.jsoup.nodes.Document document = Jsoup.parse(html);
document.outputSettings().prettyPrint(false).outline(false);
System.out.println("Example 1:");
System.out.println(document.body().html());
document.outputSettings().prettyPrint(true).outline(false);
System.out.println("Example 2:");
System.out.println(document.body().html());
document.outputSettings().prettyPrint(true).outline(true);
System.out.println("Example 3:");
System.out.println(document.body().html());
```

Results in

```
Example 1:
<div><span>1:15</span>–<span>2:15</span>&nbsp;p.m.</div>
Example 2:
<div>
 <span>1:15</span>–
 <span>2:15</span>&nbsp;p.m.
</div>
Example 3:
<div>
 <span>1:15</span>
 –
 <span>2:15</span>
 &nbsp;p.m.
</div>
```

Example 1: Please watch that there are no white spaces around the en-dash. That's correct.
Example 2: Additional white space after the en-dash.
Example 3: Additional white space before and after the en-dash.

Examples 2 and 3 result in incorrect typography and I hereby consider that as a bug.

It is likely that this problem is similar to issue #851

kovacstamasx:
Hi Edy,

I've opened a PR to fix your issue.
It would print inline tags in the same line when using pretty print without the outline option.

Though I'm not sure that your Example 3 is a bug.
When you use `outline(true)` all **tags** and **nodes** should be handled **as blocks**.
And line break is needed when you pretty print blocks.

jhy:
Thanks! I have merged the PR (and tweaked with 279fbe120c5bd4221478481b78d31ddf41475360) to fix this.

PR #1306, opened 2020-01-26
Not add line break when pretty printing inline tags

kovacstamasx:
This is a fix for issue #1305.

### DIFF
--- a/src/main/java/org/jsoup/parser/Tag.java
+++ b/src/main/java/org/jsoup/parser/Tag.java
@@ -16,7 +16,7 @@ public class Tag {
 
     private String tagName;
     private String normalName; // always the lower case version of this tag, regardless of case preservation mode
-    private boolean isBlock = true; // block or inline
+    private boolean isBlock = true; // block
     private boolean formatAsBlock = true; // should be formatted as a block
     private boolean canContainInline = true; // only pcdata if not
     private boolean empty = false; // can hold nothing; e.g. img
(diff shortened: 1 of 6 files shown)


src/main/java/org/jsoup/parser/Tag.java
```diff
@@ -122,7 +123,7 @@ public boolean canContainBlock() {
      * @return if this tag is an inline tag.
      */
     public boolean isInline() {
-        return !isBlock;
+        return inlineTag;
```
jhy: Sorry, I don't get this one. No tag is marked as an inlineTag in this impl. https://github.com/jhy/jsoup/blob/7c60ac132afc0034da18bfe969e18046994f9427/src/main/java/org/jsoup/parser/Tag.java#L285

They get marked as !block and !formatBlock. What do we need / use this field for?

kovacstamasx: The original implementation of `isInline()` returned true for non-inline tags (e.g.: `<p>`) also.
That's why I made it stricter.

In the meantime I've noticed that happened because `<p>` is included in the `formatAsInlineTags` list.
https://github.com/jhy/jsoup/blob/7c60ac132afc0034da18bfe969e18046994f9427/src/main/java/org/jsoup/parser/Tag.java#L263
Would the desired behavior be to skip line breaks for inline tags and those listed in `formatAsInlineTags`?

jhy:
Thanks, I've merged this in at 3a9bf46eeb36196d841c3fa05597e40a13f51f43, and tweaked some settings to clean up Tag, and not add line breaks for inline tags following text.

At some point we should re-do Tag and the formatters. The formatAsInline should be formatContentsAsInline or similar. Tag still contains legacy from when it was used during parsing to understand flow, which has long since been swapped out with the actual HTML Tree states. Now, it should be simplified and just used for formating.

PR #1316, opened 2020-02-06
Avoid unnecessary null checks, as Element#indexInList never returns n…

PascalSchumacher:
…ull.

Element#indexInList returns a primitive integer, so null will never be returned.


jhy:
Thanks!

PR #1317, opened 2020-02-06
Avoid unnecessary boxing of char in TokenQueue#chompBalanced.

PascalSchumacher:
TokenQueue#consume returns a primitive character, so boxing seems unnecessary.


jhy:
Thanks, merged

PR #1318, opened 2020-02-06
Use String#replace instead of String#replaceAll where possible.

PascalSchumacher:
String#replace does not use a regular expression for the replacement on Java 9+ and is therefore faster.


PascalSchumacher:
Travis build failure on Java 12 only:

>No JDK version is compatible with openjdk@1.12.. Available JDKs are:
>The command "bash .install-jdk-travis.sh" failed and exited with 2 during.

I guess the Java 12 build should be removed, as Java 12 was superceeded by Java 13 and is not supported any more.

jhy:
Thanks! Merged

PR #1323, opened 2020-02-13
Apply attribute normalization to void tags

kalaider:
Fix missing attribute normalization In the `insertEmpty` method of `HtmlTreeBuilder`.

Solves the same problem as #1224 but more naturally respecting case sensitivity settings for xml parsers.


jhy:
Thanks! Merged in 1d9f465895af9141e361b276322fc0bdb323b566 and also noticed Forms had the same issue, so I fixed that also with 6e8241ab59152f1e87a9276654c67534124793c7

PR #1335, opened 2020-03-04
Java8 and JUnit5

offa:
Update to Java 8 and JUnit 5.

While porting the tests and assertions to JUnit 5 wasn't difficult, the `@Rule`'s were tricky as they are no longer supported. I've implemented tests using the former locale rule as parametrized tests, which provide the benefit of test results for each locale. This has a drawback unfortunately: JUnit passes the parameters as argument, so the test must set the locale at the beginning – I haven't figured out how to automate this yet; the default locale is restored automatically however.

@jhy there's an unused test case ([`ElementTest::testChildSizeWithMixedContent()`](https://github.com/jhy/jsoup/blob/master/src/test/java/org/jsoup/nodes/ElementTest.java#L1599) which isn't marked with `@Test` nor used anywhere else. Is this intentional? It looks like accidental to me.


jhy:
This is great, thanks @offa. I have merged this in 8e432a5f5fde4694834ce23c5ac1503ce8d381bc. 

And also fixed that unused test case, and updated min Android API level from 8 to 10.

PR #1351, opened 2020-04-15
Add select method overloads with Evaluator.

nstdio:


jhy:
Thanks, but not seeing much value from this one-liner

PR #1354, opened 2020-04-19
Support Element.firstParent(query)

### LINKED ISSUE #1060: [Enhancement] Element.firstParent(String query)
It would be nice if there was a function that allows one to find the nearest parent using a specific query, for readability purposes.

Currently what is happening is
`Element.parents().select(query).first()`

What it would look like
`Element.firstParent(query)`

InanisV:
Fixes #1060


jhy:
Thanks - similar concept implement now for https://jsoup.org/apidocs/org/jsoup/nodes/Element.html#closest(java.lang.String)

Issue #1344, opened 2020-03-27
Behavior of `title()` on XHTML doesn't match browser behavior when `<title>` has child elements


xenomachina:
If I parse:

``` xhtml
<?xml version="1.0" encoding="UTF-8"?>
<html xmlns="http://www.w3.org/1999/xhtml">
    <head>
        <title>normal &lt;3 <script>alert()</script> <b>bold</b>!</title>
    </head>
</html>
```

with `org.jsoup.parser.Parser.xmlParser()`, the `title()` of the resulting `Document` is `"normal <3 alert() bold!"`, but if I view this document in either Firefox of Google Chrome the title is "normal <3 !".

It looks like the browsers strip out child nodes entirely, while jsoup is just stripping the start/end tags, but preserving the descendant text nodes.

Note: this only impacts XHTML. jsoup appears to match browser behavior for the equivalent HTML.

jhy:
The XML parser is not an XHTML parser - it just parses regular XML (with a few issues because jsoup was built initially with a focus on HTML). The regular HTML parser will parse XHTML.

So I consider the XML parser output correct in this instance (but different to your report - see https://try.jsoup.org/~GCS931U--xbAFD4Cid4VUEQwrys ). The HTML parser, maybe not -- it was correct per the HTML5 spec at time of implementation afaik - eat up to the `</title>` and treat it as plain text. But worth a review.

jhy:
Apologies for the late review. I double checked and believe this is working correctly. The HTML parser parses all that content within the title tag as a textnode with the literal content `normal <3 <script>alert()</script> <b>bold</b>!`. Which is how browsers do / HTML5 spec requires.

In XML parse mode, it parses the content as regular elements to a tree containing `script`, `b` elements etc. Because the `title()` method doesn't mean much in XML, the implementation is just returning the text content of the tree.

Note that XHTML should be parsed in jsoup using the HTML parser to get the HTML rules applied. The XML parser is generic and does not apply any specific HTML tree building rules. The XHTML modes in jsoup are more about how output serialization is performed.

PR #1357, opened 2020-04-20
Fix the problem that parser does not strip children nodes in the Document.title() method in xml file.

WinstonHuTiger:
#1344


jhy:
Thanks. I think the current behavior of the XML parser is correct (not treating title tags as anything specific). The title method doesn't make a lot of sense outside of HTML documents, but I think gathering the text is a suitable approach. So, I don't see a need to change the behaviour.

PR #1359, opened 2020-04-20
Add a parse method to encode illegal tag in html body

### LINKED ISSUE #1230: How to encode illegal tag in html body
Is there anyway to encode illegal tags with cleaner or parser like this:

input: `<p>Try <aaa>jsoup</p>`
output: `<p>Try &lt;aaa&gt; jsoup</p>`

thanks for your help.

InanisV:
Fixes #1230

### DIFF
--- a/src/main/java/org/jsoup/Jsoup.java
+++ b/src/main/java/org/jsoup/Jsoup.java
@@ -59,6 +59,19 @@ public class Jsoup {
     }
 
     /**
+     Parse HTML into a Document with only legal tags by encoding illegal ones.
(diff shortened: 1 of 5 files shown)


jhy:
Thanks, but no thanks. Sorry for the late reply. A few thoughts:

If we were to add this functionality, I think it makes more sense as an option to the Cleaner - what to do on an error - drop or escape.

The internal use of reflection strikes me as an odd pattern here. If we wanted to access these fields it would be better to make public methods, and not bind it completely to the current implementation.

The naming of things as 'legal' and 'illegal' doesn't really make sense to me. All of these tags are valid HTML. Maybe known/unknown would be better.

Issue #764, opened 2016-09-27
jsoup will not treat an incoming <Alpha as text


xenoterracide:
Jsoup does not treat this as text

```
 Jsoup.parseBodyFragment("<mg per week or  )" );
```

it seems to just discard this entirely, and I don't see a way around it, I was going to try to get jsoup to encode with &lt, but the parser is just discarding it. Also `isValid` returns true so I'm not sure how I could handle the problem on the fly. I've stepped through the debbuger, there are no children/attributes

```
<p><mg per week or  )</p>
```

becomes

```
<p><mg per week or )<></mg></p>
```

and this (and `<div>` )

```
<span><mg per week or  )</span>
```

becomes

```
<span><mg per week or )< pan></mg></span>
```

version 1.9.2

ksclarke:
I think my issue may be similar to this one? If I try to JSoup clean `If A<B then B>A` I get `If A<b>A</b>`. I don't think element names are allowed to have spaces in them, but JSoup is parsing this text as an element.

xenoterracide:
You know, I'd complain about the resolution, because really I'm not sure how I could have encoded this garbage correctly (only from rereading this issue), yes I know it was garbage input, but that was the problem. I mean, how does the last example become `< pan>`, in what world does that make any sense. Really, I'm just going to complain that this wasn't responded to for *5 YEARS*.

jhy:
To be specific, it gets parsed like:

In:
```
<span><mg per week or  )</span>
```

Out DOM (with verbose attributes for clarity):
```
<span>
  <mg per="" week="" or="" )>="" span=""></mg>
</span>
```

And how chromium sees it (same):
![image](https://user-images.githubusercontent.com/76934/131953673-f030502e-4f2c-407d-bf33-823750f9ddce.png)


Which makes sense to me as far as it goes, given the constraints of HTML supporting custom tags and boolean attributes.

xenoterracide:
gross, and still doesn't make sense to me, I wonder how chromium is getting that too. I mean, I understand the `<mg></mg>` just not how it's creating `< span>`, that seems like total garbage (though so does my input), and not sure why it was `< pan>` 60% certain that's a copy/paste.

jhy:
There is no `< span>` element. As I wrote, that is an attribute name. If you are actually interested in how both of these parsers are getting this as the output, it is covered in the InBody section of the TreeBuilder algorithm in the WhatWG spec.

PR #1362, opened 2020-04-27
Fix the problem that jsoup will not treat an incoming <Alpha as text

Tricker-z:
Fix issue #764 
## Proposed changes
~~~
Jsoup.parseBodyFragment("<mg per week or  )" );
~~~
becomes
~~~
&lt;mg per week or )
~~~
and
~~~
Jsoup.parseBodyFragment("<p><mg per week or )</p>" );
~~~
becomes
~~~
<p>&lt;mg per week or )</p>
~~~

## Checklist
- [ ] Unit tests are passed locally after my changes
- [ ] I have added test to prove the validity

### DIFF
--- a/src/main/java/org/jsoup/parser/CharacterReader.java
+++ b/src/main/java/org/jsoup/parser/CharacterReader.java
@@ -10,7 +10,7 @@ import java.util.Arrays;
 import java.util.Locale;
 
 /**
- CharacterReader consumes tokens off a string. Used internally by jsoup. API subject to changes.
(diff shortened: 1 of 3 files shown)


jhy:
Thanks for the PR. I don't want to move jsoup away from the HTML5 spec in this way. We actually previously supported it but deliberately aligned to the spec to reduce parse errors people experienced.

This PR reformats much of the file and so the diff is hard to read. From what I can make out, I expect that this will bug out during bufferUp() hits.

PR #1370, opened 2020-05-08
Issue 887 has been fixed.

HollowKnight123:
Invalid characters in XML(e.g. &#11, &#12) have been special judged and removed during parsing.
Only valid XML characters(e.g. &#9, &#13) will show up in the document as result.


jhy:
Thanks and sorry for the late reply. The underlying issue was fixed in for #1556. That discards illegal XML characters on serialization, vs in parsing.

PR #1371, opened 2020-05-08
basic fix for Issue 1117

PneuC:
I tried a basic fix for issue 1117. The down level revealed will be parse into a whole Comment node with an additional atrribute "isDownLevelRevealed=true". And now both the result of token.toString and document.html will show down level revealed like "<!if true>1<!endif>" but not " 1" any more.


jhy:
Thanks, and sorry for the late reply. I currently don't plan on integrating this change, particularly as the parse feature is non standardized.

Issue #1361, opened 2020-04-23
HTML-Tag has now 4 childs instead of 3 as in 12.2


leomayer:
I have the following content

```
<html> 
 <head> 
  <title>Example - Text Formating Tags</title> 
 </head> 
 <body> 
  test
 </body>
</html>
```
when I traverse the document with 

			final Document doc = Jsoup.parse(htmlText);
			NodeTraversor.traverse(myRoot, doc);

I get an empty `TextNode` just before the `head` tag. Up till 12.2 there wasn't any additonal `TextNode`. Can this be fixed to have this additonal `TextNode` be omitted?

The `output.prettyprint` has as expected no influence on this behaviour.

StarPolygen:
Could you give test input and output more clearly?

StarPolygen:
Ok, This cannot repo in version 1.13.1 [2020-Feb-29]:
>
String html = "<html><head><title>Example - Text Formating Tags</title></head><body>test</body></html>";
        final StringBuilder accum = new StringBuilder();
        final Document doc = Jsoup.parse(html);
        NodeTraversor.traverse(new NodeVisitor() {
            @Override
            public void head(Node node, int depth) {
                accum.append("<").append(node.nodeName()).append(">");
                return;
            }

            @Override
            public void tail(Node node, int depth) {
                accum.append("</").append(node.nodeName()).append(">");
                return;
            }
        }, doc);
        System.out.println(accum.toString());

And the result is as follows:
>
<#document><html><head><title><#text></#text></title></head><body><#text></#text></body></html></#document>

The empty node did not appeared.

StarPolygen:
Or It may cause by "\n":
For example:
```
String html = "<html>\n" +
                "<head>\n" +
                "<title>Example - Text Formating Tags</title>\n" +
                "</head>\n" +
                "<body>\n" +
                "test\n" +
                "</body>\n" +
                "</html>";
```
The traverse result can be: <#document><html><#text></#text><head><#text></#text><title><#text></#text></title><#text></#text></head><#text></#text><body><#text></#text><#text></#text></body></html></#document>
with the same operation I said above.

StarPolygen:
I think It can be quicker if we can preprocess "\n" such as ignore it in Html text. And maybe some user want to ignore specified context in HTML text. So I add a class to jsoup to solve the prob.

leomayer:
[...]

PR #1372, opened 2020-05-08
Fix Issue #1361 may bring higher performance

StarPolygen:
Add  Class HtmlShrinker(org.jsoup.parser.HtmlShrinker)  and its corresponding test(org.jsoup.parser.ShrinkTest):

Issue #1361 shows parser generate spare empty TextNode for "\n".
Ignore "\n" can reduce amount of nodes, thus increase performance each time traverse the Tree.

To be extend, I add a class HtmlShrinker used to ignore any content string specified by users. Not only "\n", this feature may help users to filter sometext they think useless in html text.

### DIFF
--- a/src/main/java/org/jsoup/parser/HtmlShrinker.java
+++ b/src/main/java/org/jsoup/parser/HtmlShrinker.java
@@ -1,0 +1,83 @@
+package org.jsoup.parser;
+
+import java.util.Vector;
+
+/**
+ Increase efficiency: Shrink html text by ignoring specified content.
+ An additional feature provided to users.
+ Can reduce amount of Textnode by ignore "\n" to speed up parse.*/
+public class HtmlShrinker {
+
+    private static int[] getNext(String aim) {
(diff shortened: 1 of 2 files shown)


StarPolygen:
Add: The implement has considered the influence of escape characters

SliverySky:
I agree with you

leomayer:
As a quick remark to this fix: Shouldn't be `\t` be treated the same way as `\n`? I know for formatting it would be great to have them - but traversing I don't see any benefit at all. 
The q arose due to the test cases which are only for `\n`. If you don't think `\t` should be treated similar - just ignore my comment.

StarPolygen:
OK， Sorry for late. I think your concern is reasonable. Here I did not test the cases for "\t". Just use the original testcase at #1361. If you  want to treat "\t" or any other string as I treat in my testcase. Maybe you can check the method ""

jhy:
Thanks, but I don't see much call for this, and no measurement of any performance increase vs the cost of inclusion.

PR #1374, opened 2020-05-08
Fix the bug in issue 845

SliverySky:
I check how chrome convert the html which have three nested <a> tag, it is very strange. Since <a><a></a></a> is illegal, the result is unpredictable. It is better to write a method to reconstruct the dom tree to avoid nested <a> tag, rahter than modify the process of constructing dom tree.  So I write a class to handle illegal dom tree. However, there is only one method which handle nested <a> tag in this class now. I also add some test case.


jhy:
Thanks - went with another PR #1517 that implements the spec's adoption agency algo

PR #1375, opened 2020-05-08
I fix the bug in 986 and add corresponding test.

SliverySky:
I fix the bug in 986 and add corresponding test. I modify the chompBalance() by ignoring the brackets which are between \Q \E. Then it work well.


jhy:
Thanks - took a similar PR in for the same issue - d4f91c5eba9be87cd7d6ca90275e2200a4a6ff65

PR #1381, opened 2020-05-19
Update dependencies

edumco:
### Several dependencies has been upgraded on master

In this PR I added the @dependabot suggestions one by one until all are passing on CI


jhy:
Thanks and sorry for the late reply - have applied similar updates recently

Issue #876, opened 2017-04-27
Should normalize the input text in :contains() queries


unitydynamics:
I've been a JSoup user for a long time.  I appreciate the work that went into it!

I've found a circumstance which tripped me up.  I was using the :contains(With  Text)

(that's 2 spaces between "With" and "Text".  The raw HTML has the text this way.  But it seems JSoup matches on the normalized text.  If this is the expected behavior maybe just add a note to the documentation on it?  Thanks.  Or even better... can we use quotes to force matching against the un-normalized text?  I suppose it might be confusing if  :contains(With  Text) and  :contains("With  Text") were different.

jhy:
It seems like it would be better to normalize the `contains` text query so we're comparing apples to apples.

Ruefors:
I wonder if this is a formatting issue because the parser will merge extra Spaces when parsing,You also need to format a string when using the Cotaines search, such as merging extra Spaces

jhy:
Taking another look at this and somewhat related #576. I think we can address this, make some improvements, and also maintain backwards compatibility:

 - for `:contains(text)`, normalize the selector text in the same way that the `text() ` and `ownText()` methods work. That directly fixes this issue. I conceptually like the neatness of using `:contains("text")` to use a non-normalized search, but that's a bit stringly typed, and maybe backwards in-compat.
 - introduce something like `:containsWhole` and :containsOwnWhole which uses `wholeText()` and no normalization to either the input or the query. That way we can find content including newlines, tabs, repeated spaces, etc
 - for #576, introduce `:matchesWhole` and `:matchesOwnWhole`, to work similarly for regular expressions

PR #1383, opened 2020-05-21
Fix for issue #876

michelpetrovic:
https://github.com/jhy/jsoup/issues/876

Added a compiled RegEx, for maximum performance, for checking and removing multiple spaces.

Though not directly related to the issue, I also modified `Normalizer.normalize(String)`; by doing `trim()` before `toLowerCase()`, it means potentially less characters to iterate over in `toLowerCase()`.


jhy:
Sorry for the late review. I fixed the underlying issue #876 a whiles back.

Issue #1289, opened 2019-12-20
NPE after replaceWith in NodeVisitor


codemonstur:
Jsoup throws an NPE when traversing a DOM with a NodeVisitor after a replaceWith call in the head(). Example code:
```
    private static final String html =
        "<html>\n" +
        "<head>\n" +
        "    <link href=\"css/test.css\" rel=\"stylesheet\">\n" +
        "</head>\n" +
        "<body>\n" +
        "</body>\n" +
        "</html>";
    public static void main(final String... args) {
        Document document = Jsoup.parse(html);
        document.traverse(new NodeVisitor() {
            @Override
            public void head(Node node, int depth) {
                if (node instanceof Element) {
                    Element element = (Element) node;
                    if ("link".equals(element.tagName())) {
                        Element style = element.ownerDocument().createElement("style");
                        element.replaceWith(style);
                    }
                }
            }

            @Override
            public void tail(Node node, int depth) {}
        });
    }
```
I guess this makes sense. But I feel this could be considered a bug. Perhaps the code should simply move on to the next sibling or the parent. Perhaps it should throw a DontReplaceWhatImUsingException.

jhy:
Thanks - have implemented support for this in the head() method. See d66decd3c6ce450c19f276e9693a2727b7aa9748 for a test case example.

PR #1386, opened 2020-05-28
add a new Exception for NPE after replaceWith in Node visitor

HcPlu:
#1289


jhy:
Thanks -- lots of changes here that seem outside of the scope of the original issue. I implemented support for that in d66decd3c6ce450c19f276e9693a2727b7aa9748

Issue #953, opened 2017-10-15
Add method Elements.before(Node node)


AnaelMobilia:
Hello,

We can made `Element.before(Node node)`, `Element.before(String html)` and `Elements.before(String html)` .

Is it possible to add `Elements.before(Node node)` ?
Actually, I need to do `myElements.before(myElement.outerHtml())` which isn't really pretty...

Regards,
Anael

jhy:
Hi -- can you describe your use case or show an example? Just want to make sure I understand the request completely.

AnaelMobilia:
Hello,

I found an element an the webpage. (eg "<p>Warning : remote link</p>")
I want to insert it before all div[class=specific]  elements.

I need to output my element to HTML to add it before each selected div.
IMHO, it could be great to insert it as an element (and don't needing to put it on html)

Regards,
Anael

HcPlu:
I am trying to do like`Elements` to solve it

jhy:
Should do this for before, after, prepend, append, to match the Elements string methods.

AzazelSensei:
Opened https://github.com/jhy/jsoup/pull/2567 for this. before/after/prepend/append(Node) on Elements, clone per target.

PR #1387, opened 2020-05-28
[Feature]Add  method Elements.Befor(Node node)

HcPlu:
#953


jhy:
Thanks, but most of this diff is irrelevant and duplicating the Cleaner about the place (?). Not appropriate to merge.

Issue #1294, opened 2020-01-07
The ruby rtc element is incorrectly processed.


rhdunn:
Given the markup in the example from https://www.w3.org/TR/2001/REC-ruby-20010531/#complex:
```
<ruby>
  <rbc>
    <rb>10</rb>
    <rb>31</rb>
    <rb>2002</rb>
  </rbc>
  <rtc>
    <rt>Month</rt>
    <rt>Day</rt>
    <rt>Year</rt>
  </rtc>
  <rtc>
    <rt rbspan="3">Expiration Date</rt>
  </rtc>
</ruby>
```

the jsoup parser treats the `rtc` element as an unknown element that gets closed immediately. This causes it to serialize in xml mode as:

    <rtc></rtc><rt>Month</rt><rt>Day</rt><rt>Year</rt>

I have checked the behaviour of Firefox and Chrome, and they preserve the `rtc` element structure, e.g.:

    <rtc><rt>Month</rt><rt>Day</rt><rt>Year</rt></rtc>

The `rtc` element is supported in the W3C HTML spec [1], but not the WHATWG spec. Also, even though the `rbc` element is not listed in either of those (only in the Ruby Annotations specification), the jsoup parser preserves the `rbc` element structure.

[1] https://www.w3.org/TR/2014/REC-html5-20141028/text-level-semantics.html#the-rtc-element

jhy:
Thanks -- fixed. I brought the implementation up to the current spec as defined by [WHATWG](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inbody) (scroll to 'A start tag whose tag name is one of: "rb", "rtc"' and the following "rp" / "rt" section).

Note that in jsoup (and other browsers), tags which aren't explicitly defined are still supported, they get default treatment.  The bug described above was not because "rtc" was not known, it was because there was explicit handling for it -- and the spec changed since implemented.

I'd appreciate it if interested users can test and review, and raise any issues found.

jhy:
Also, I checked that Tag defines the tags `ruby`, `rp`, `rt`, which are the only elements defined in the element spec - https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-ruby-element

Those are all marked as inline (phrasing). Other tags (`rtc`, `rc`) are not explicitly defined and so get treated as block. If those are used commonly in the wild, we could add those to Tag as inline to help formatting.

PR #1389, opened 2020-05-28
fix issue 1294

wudiiv11:
#1294


jhy:
This is not production ready.

Issue #1006, opened 2018-01-16
Should enforce that tags start with ascii characters only


jackila:
When I use Jsoup.parse('') to parse the html 

`<p>5.到推机制，推荐会员匹配成功超过24小时不打款帐号会被冻结，扣领导人`
`<会员挂单金额5%>动态做为惩罚</p>`
`<p><br></p>`

The result come out like this:

`<p>5.到推机制，推荐会员匹配成功超过24小时不打款帐号会被冻结，扣领导人`
  `<会员挂单金额5%>`
    `动态做为惩罚`
   `</会员挂单金额5%></p>`
  `<p><br /></p>`

apparently <会员挂单金额5%> is not a tag ,is there any method  can  solve it?

krystiangorecki:
If that's the only problematic tag you can escape it manually:
``` 
html = html.replaceAll("<会员挂单金额5%>", "&lt;会员挂单金额5%&gt;");

**Comment by SliverySky** [comment]
In this version, Chinese character is also considered as legal tag. e.g<一></一> this is not allow in chrome. String "<一>"will be mistaken as unclosed tag and be replace by <一></一>. The reason is that Character.isLetter() will consider all character in all language as letter.[https://github.com/jhy/jsoup/pull/1390](url)

**Comment by jhy** [comment]
Per the HTML spec, and in checking in Chrome: tags must start with an ascii character, and any character after that is allowed.

See [tag open state](https://html.spec.whatwg.org/multipage/parsing.html#tag-open-state) followed by [tag name](https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state) state.

So, valid tags should be:

```html
<a一>
<a会员挂单金额5>
<table(╯°□°)╯>
```

Invalid:
```
<一>
<会员挂单金额5>
<table(╯°□°)╯>
```

SliverySky:
font{
        line-height: 1.6;
    }
    ul,ol{
        padding-left: 20px;
        list-style-position: inside;
    }


    
    
        Thank you for your replay which let me know the reason for this problem (why <一> can't work in Chrome). And I will standardize the format of PRs. Thank you for reminder!
            
        
    
    
        
            
        
    
    
     
    
        
                    
            
                
                    
                            
                                
                            
                            
                                11710101
                            
                    
                        
                            
                                    11710101@mail.sustc.edu.cn
                            
                        
                
            
        
        
    
    
        签名由
        网易邮箱大师
        定制
    
 

[...]

PR #1390, opened 2020-05-29
Fix the issue 1006

SliverySky:
Fix the issue 1006. In old version, Chinese character is also considered as legal tag. e.g<一></一> this is not allow in chrome. String "<一>"will be mistaken as unclosed tag and be replace by <一></一>. The reason is that Character.isLetter() will consider all character in all language as letter.

### DIFF
--- a/src/main/java/org/jsoup/parser/CharacterReader.java
+++ b/src/main/java/org/jsoup/parser/CharacterReader.java
@@ -505,10 +505,20 @@ public final class CharacterReader {
         return !isEmpty() && Arrays.binarySearch(seq, charBuf[bufPos]) >= 0;
     }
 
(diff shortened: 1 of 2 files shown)


jhy:
Thanks for the PR, and sorry for the late reply.

I checked the HTML syntax spec and see that tags must start with an ASCII character. So the update should probably just be to do that. There's nothing Chinese-lang specific here so the change should not incorporate that.

Also for style of PRs - please make sure the title indicates the specific change (I can't remember bug numbers). And add the test case to an existing testcase file.

Will close this, but please feel welcome to re-open this or a new one to confirm to the spec (as detailed in #1006).

Also I believe the `.isLetter` check was a bugfix for another issue, so we will need to review that also.

PR #1391, opened 2020-05-30
Fix incorrect template element xml serialization

### LINKED ISSUE #1315: template element xml serialization removes the inner start/end tags
After parsing the HTML fragment:

    <template id="lorem-ipsum">
      <tr>
        <td>Lorem</td>
        <td>Ipsum</td>
      </tr>
    </template>

using the following:

    doc.outputSettings().prettyPrint(false)
    doc.outputSettings().syntax(Document.OutputSettings.Syntax.xml)
    return doc.outerHtml()

the output is as follows:

    <template id="lorem-ipsum">
      
        Lorem
        Ipsum
      
    </template>

That is, it is the same as the template content excluding the start and end tags. The whitespace is still preserved.

The tree construction in Firefox and Chrome is the same as the source, with a `#document-fragment` node between the template element and the tr content. That is:

    <template id="lorem-ipsum">
      #document-fragment
        <tr>
          <td>Lorem</td>
          <td>Ipsum</td>
        </tr>
    </template>

Tricker-z:
Fix #1315 

After parsing the HTML fragment, the inner start/end tags is included.

### DIFF
--- a/src/main/java/org/jsoup/parser/TemplateTest.java
+++ b/src/main/java/org/jsoup/parser/TemplateTest.java
@@ -1,0 +1,71 @@
+package org.jsoup.parser;
+
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TemplateTest {
+    @Test
+    public void testParseTemplateWithTrTd() {
+        String html = "<template id=\"lorem-ipsum\">\n" +
+                " <tr>\n" +
+                "    <td>Lorem</td>\n" +
+                "    <td>Ipsum</td>\n" +
+                " </tr>\n" +
+                "</template>";
+
+
+        Document doc = Jsoup.parse(html);
+        doc.outputSettings().prettyPrint(false);
+        doc.outputSettings().syntax(Document.OutputSettings.Syntax.xml);
+
+
+        assertEquals("<html><head></head><body><template id=\"lorem-ipsum\">\n" +
+                " <tr>\n" +
+                "    <td>Lorem</td>\n" +
+                "    <td>Ipsum</td>\n" +
+                " </tr>\n" +
+                "</template></body></html>", doc.outerHtml());
+        System.out.println(doc.outerHtml());
+    }
+
+    @Test
+    public void testParseTemplateWidthTd() {
+        String html = "<template id=\"lorem-ipsum\">\n" +
+                "    <td>Lorem</td>\n" +
+                "    <td>Ipsum</td>\n" +
+                "</template>";
+
+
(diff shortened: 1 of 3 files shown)


src/test/java/org/jsoup/parser/TemplateTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+        Document doc = Jsoup.parse(html);
+        doc.outputSettings().prettyPrint(false);
+        doc.outputSettings().syntax(Document.OutputSettings.Syntax.xml);
+
+        assertEquals("<html><head></head><body><template id=\"lorem-ipsum\">\n" +
+                "    <td>Lorem</td>\n" +
+                "    <td>Ipsum</td>\n" +
+                "</template></body></html>", doc.outerHtml());
+        System.out.println(doc.outerHtml());
```
bartektartanus, line 26: All `println` should be removed from tests.

src/test/java/org/jsoup/parser/TemplateTest.java
```diff
@@ -0,0 +1,60 @@
...
+                "    <td>Lorem</td>\n" +
+                "    <td>Ipsum</td>\n" +
+                "</template>";
+
+
+        Document doc = Jsoup.parse(html);
+        doc.outputSettings().prettyPrint(false);
+        doc.outputSettings().syntax(Document.OutputSettings.Syntax.xml);
+
+        assertEquals("<html><head></head><body><template id=\"lorem-ipsum\">\n" +
```
bartektartanus, line 22: Maybe instead of checking HTML literally you should verify that `doc.body()` is `template` tag and has two `td` tags as children?

src/main/java/org/jsoup/parser/TemplateTest.java
```diff
@@ -0,0 +1,71 @@
+package org.jsoup.parser;
```
bartektartanus, line 1: This test is duplicated and also added to `src` instead of `test`. Remove this.

src/test/java/org/jsoup/parser/TemplateTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+
+        assertEquals("<html><head></head><body><template id=\"lorem-ipsum\">\n" +
+                "    <tr>Lorem</tr>\n" +
+                "</template></body></html>", doc.outerHtml());
+        System.out.println(doc.outerHtml());
+    }
+
+    @Test
+    public void parseXml() {
```
bartektartanus, line 48: This test does not verify anything. Add some assertions or remove it.

bartektartanus (changes requested):

jhy:
Thanks for this, and sorry for the late reply! I have implemented a similar change, aligned to the HTML5 spec, in #1634.

Issue #784, opened 2016-11-03
Add an abs: pseudo-class selector


jhy:
We should have an `abs:` pseudo-class selector, so you can select on the absolute URL of an element. I.e. `a[abs:href=...]`, `img[abs:src$=...]` etc. Now it could only be on the specific href attribute value.

Also could think about doing URL parsing and then able to select on host, path, query, spec, et all.

Like `a[host:href=example.com]`. You can kind of do that now with the range query, but that's kind of susceptible to input variations.

jhy:
Closing; better done in code than in a selector.

PR #1392, opened 2020-05-30
Jsoup issue#784 commit

ChrisZcu:
commit the issue#784 fixed


ChrisZcu:
jsoup issue #784 fix

jhy:
Hi, thanks for this and sorry for the late reply. This looks interesting, but isn't really in the style of the rest of jsoup, and creates some selectors that are going to be confusing against attribute names. Will close, but feel free to discuss on the original issue or in another discussion more about the proposal and we can align on the new selectors and the implementation.

PR #1394, opened 2020-05-30
Fix NullPointerException on standalone elements

### LINKED ISSUE #593: Node#wrap() throws NullPointerException on standalone elements
When trying to call wrap(String) on a standalone element, wrap accesses the parent without checking if it is null which causes an exception.

Example

```
Document doc = Jsoup.connect('http://www.google.com').get();
// create a new standalone element
Element standaloneElement = new Element(doc.tag(), doc.baseUri())
// wrap accesses parent which is null
standaloneElement.wrap("<div></div>")
```

And the exception

```
java.lang.NullPointerException
    at org.jsoup.nodes.Node.wrap(Node.java:360)
    at org.jsoup.nodes.Element.wrap(Element.java:453)
```

Tested on version 1.8.2

Tricker-z:
Fix #593 

standalone element can wrap tags correctly

### DIFF
--- a/src/main/java/org/jsoup/nodes/Node.java
+++ b/src/main/java/org/jsoup/nodes/Node.java
@@ -356,16 +356,28 @@ public abstract class Node implements Cloneable {
 
         Element wrap = (Element) wrapNode;
         Element deepest = getDeepChild(wrap);
-        parentNode.replaceChild(this, wrap);
-        deepest.addChildren(this);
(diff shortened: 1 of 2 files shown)


jhy:
Thanks - same issue was fixed with 77fcaf4dcdc8fd185e2f5d8eb6aa9c425a38bc9b

Issue #1173, opened 2019-01-13
Add support for ::before and ::after selectors 


aleksi1:
::before and ::after selectors are widely used and supported, can you add support for these selectors?

Example:

`    	Elements e = Jsoup.parse("<span><div>").select("div::before");
`

https://caniuse.com/#feat=css-gencontent

Seeyn:
can you give a more specific example?

Seeyn:
Like:
<div>
 a<span>b</span><span>c</span>
</div>

.select("span:before")

aleksi1:
I think that in browser there is no real element for ::before selection, but you can add CSS styles to that selection. I think it could work as before / after methods in Element. So kind of invisible element, which you could convert to real element by adding html / tag etc to it.

Seeyn:
Thanks for your reply.So, your intention is to convert the invisible element to real element according to the css style. am i right?

aleksi1:
My use case is that I cannot use Element API and my app user can only use CSS selectors, so it would be good if I could somehow use ::before / ::after selectors and append HTML to selected position just like in browser.

jhy:
Thanks for the discussion - I'm not really understanding the use-case or the proposal. How are you going to append HTML to selected items if you can't use the API? Why can't you use the API?

Happy to review with more of a specific use case and a proposal.

PR #1396, opened 2020-05-30
jsoup issue #1173 fix code commit

ChrisZcu:
Jsoup #1173 fix code commit


jhy:
Thanks - if you want to proceed with this, could you submit the PR with a descriptive subject and comment? E.g. "Adds support for ::before and ::after". And, the diff should only include your changed lines; the reformatting you did has changed many lines and makes the change hard to audit. 

Same comments apply to your other PRs.

jhy:
Also, the commit doesn't actually do anything. You are just eating the strings "::before" or "::after". Really not understanding this at all...

Issue #1352, opened 2020-04-17
Performance issue in StringUtil


utchibutchi:
Hi, we're having performance issues with the jsoup lib. We are working on servers with very many cores and want to parse html highly parallel. Unfortunatelly there is a border where we can not increase throughput anymore by increasing threads.

I created a thread dump and nearly all threads are waiting here:

``
INFO   | jvm 1    | 2020/04/17 12:38:23 | "thread-55" ...  waiting for monitor entry
INFO   | jvm 1    | 2020/04/17 12:38:23 |    java.lang.Thread.State: BLOCKED (on object monitor)
INFO   | jvm 1    | 2020/04/17 12:38:23 |       at org.jsoup.internal.StringUtil.borrowBuilder(StringUtil.java:243)
INFO   | jvm 1    | 2020/04/17 12:38:23 |       - waiting to lock <0x00007ef4a2361248> (a java.util.Stack)
INFO   | jvm 1    | 2020/04/17 12:38:23 |       at org.jsoup.internal.StringUtil.normaliseWhitespace(StringUtil.java:146)
INFO   | jvm 1    | 2020/04/17 12:38:23 |       at org.jsoup.nodes.TextNode.text(TextNode.java:46)
``

WinstonHuTiger:
How did you check the thread usage? I mean what tools you used in your project?

WinstonHuTiger:
After we dig into the StringUtil class, we found that after increasing the size of memory, the performance will increase while the number of threads increases.

utchibutchi:
Yes, that's true until a certain point it get's better. But as said, we have very big machines, in detail they have 80 cpu cores, 196GB Memory and a threadpool with the size of up to 200 was used. In the thread dump we could see that nearly all of them were waiting for the monitor as shown above.

As a workaround we now re-implemented the TextNode.text() funktion like this:
```
String wholeText = node.getWholeText();
StringBuilder sb = new StringBuilder();
StringUtil.appendNormalisedWhitespace(sb, wholeText, false);
return sb.toString();
```

To not use a StringBuilder from the pool and it clearly solved the performance problem and we don't see problems related to the many StringBuilders created.

jhy:
Thanks for the report. This is fixed by #1402.

PR #1402, opened 2020-06-15
improve StringBuilder cache performance under high concurrency

morokosi:
Use ThreadLocal stack to store StringBuilder instance cache to avoid thread contention introduced by synchronized

Our application which utilizes multithreads to parse html documents experienced performance degradation after updating from jsoup-1.10.x to recent version. Thread dump shows threads stuck at StringBuilder.borrowBuilder / releaseBuilder so the situation is very similar to #1352. It seems synchronized blocks in borrowBuilder / releaseBuilder is introduced in https://github.com/jhy/jsoup/commit/c8c05694da5489221c27e45a5e49d0f5fcb41863.

I conducted a microbenchmark of the improvement, which shows ~10x reduction in multithreaded StringBuilder.borrowBuilder / releaseBuilder operations.
https://github.com/morokosi/jsoup/pull/1

### DIFF
--- a/src/main/java/org/jsoup/internal/StringUtil.java
+++ b/src/main/java/org/jsoup/internal/StringUtil.java
@@ -229,7 +229,12 @@ public final class StringUtil {
         }
     }
 
-    private static final Stack<StringBuilder> builders = new Stack<>();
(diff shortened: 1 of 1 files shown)


jhy:
Thanks, merged! Really appreciate the analysis and performance test.

Issue #1384, opened 2020-05-26
problems with forms()


marcoscale98:
Elements.forms() don't take any forms. 
Try to run example 3 https://www.programcreek.com/java-api-examples/?api=org.jsoup.nodes.FormElement

sham0688:
`Jsoup.parse(form_smaple).select("form").forms()` is empty on v1.13.1 (v1.11.3 OK)
[form_smaple.txt](https://github.com/jhy/jsoup/files/4752091/form_smaple.txt)

apflieger:
A backward compatibility break has been introduced here https://github.com/jhy/jsoup/commit/a657ae0240b875a703a1e4909d56ad59997d362d#diff-4f0778cbabadac4729bae10ebec0c0aaL636

the `forms()` method now return forms from child node only. This is a bug because the code is written to keep the previous behavior as it was. I'll try to make a PR

marcoscale98:
Sorry, how do I download the updated stable version? On the site I still find 1.13.1

apflieger:
I just submitted a pull request. I'm waiting on a feedback from a maintainer (@jhy ? 😄 ). It's not available yet

apflieger:
Well, my PR is still pending 😢

jhy:
Thanks, have applied the PR. 

It's unfortunate at I did have a testcase for this (the sample code referred to by @marcoscale98 is actually from jsoup, not sure why Program Creek is attributing it to another project and GPL when it's MIT). However it is in a disabled test-class as it was relying on an external server from tests I'd written before adding a local server. Need to spend some time updating those.

https://github.com/jhy/jsoup/blob/8e432a5f5fde4694834ce23c5ac1503ce8d381bc/src/test/java/org/jsoup/integration/UrlConnectTest.java#L331

PR #1403, opened 2020-06-15
Fix regression on Elements.forms()

apflieger:
the `Elements.forms()` method used to return FormElement-s that are directly in the Elements collection. This behaviour was broken since a657ae0240b875a703a1e4909d56ad59997d362d. See https://github.com/jhy/jsoup/issues/1384

I wrote the fix accordingly to what's written in the javadocs of the methods involved:
* `comments`, `textNodes` and `dataNodes` says they return direct child nodes. I specialized the private `nodesOfType` method to handle child node lookup only (and renamed it `childNodesOfType`)
* `forms()` doc says it returns the forms elements present in the selected elements, not child nodes. I removed the use of `nodesOfType` and put back the previous code.

### DIFF
--- a/src/main/java/org/jsoup/select/Elements.java
+++ b/src/main/java/org/jsoup/select/Elements.java
@@ -636,7 +636,11 @@ public class Elements extends ArrayList<Element> {
      * no forms.
      */
     public List<FormElement> forms() {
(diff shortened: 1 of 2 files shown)


jhy:
Thanks! Merged.

PR #1408, opened 2020-06-25
Rename whitelist to allowlist

drei01:
This PR renames the `Whitelist` class to `Allowlist` to make the language more welcoming and inclusive. This follows the [IETF RFC on "Terminology, Power and Offensive Language"](https://tools.ietf.org/id/draft-knodel-terminology-01.html#rfc.section.1.2).

### DIFF
--- a/src/main/java/org/jsoup/Jsoup.java
+++ b/src/main/java/org/jsoup/Jsoup.java
@@ -3,7 +3,7 @@ package org.jsoup;
 import org.jsoup.nodes.Document;
 import org.jsoup.parser.Parser;
 import org.jsoup.safety.Cleaner;
-import org.jsoup.safety.Whitelist;
+import org.jsoup.safety.Allowlist;
 import org.jsoup.helper.DataUtil;
 import org.jsoup.helper.HttpConnection;
 
@@ -189,14 +189,14 @@ public class Jsoup {
 
      @param bodyHtml  input untrusted HTML (body fragment)
      @param baseUri   URL to resolve relative URLs against
-     @param whitelist white-list of permitted HTML elements
+     @param allowlist white-list of permitted HTML elements
      @return safe HTML (body fragment)
 
      @see Cleaner#clean(Document)
      */
-    public static String clean(String bodyHtml, String baseUri, Whitelist whitelist) {
+    public static String clean(String bodyHtml, String baseUri, Allowlist allowlist) {
         Document dirty = parseBodyFragment(bodyHtml, baseUri);
-        Cleaner cleaner = new Cleaner(whitelist);
+        Cleaner cleaner = new Cleaner(allowlist);
         Document clean = cleaner.clean(dirty);
         return clean.body().html();
     }
@@ -206,13 +206,13 @@ public class Jsoup {
      tags and attributes.
 
      @param bodyHtml  input untrusted HTML (body fragment)
-     @param whitelist white-list of permitted HTML elements
+     @param allowlist white-list of permitted HTML elements
      @return safe HTML (body fragment)
 
      @see Cleaner#clean(Document)
      */
-    public static String clean(String bodyHtml, Whitelist whitelist) {
-        return clean(bodyHtml, "", whitelist);
+    public static String clean(String bodyHtml, Allowlist allowlist) {
+        return clean(bodyHtml, "", allowlist);
     }
 
     /**
@@ -224,14 +224,14 @@ public class Jsoup {
      *
      * @param bodyHtml input untrusted HTML (body fragment)
      * @param baseUri URL to resolve relative URLs against
-     * @param whitelist white-list of permitted HTML elements
+     * @param allowlist white-list of permitted HTML elements
      * @param outputSettings document output settings; use to control pretty-printing and entity escape modes
      * @return safe HTML (body fragment)
      * @see Cleaner#clean(Document)
      */
-    public static String clean(String bodyHtml, String baseUri, Whitelist whitelist, Document.OutputSettings outputSettings) {
+    public static String clean(String bodyHtml, String baseUri, Allowlist allowlist, Document.OutputSettings outputSettings) {
         Document dirty = parseBodyFragment(bodyHtml, baseUri);
-        Cleaner cleaner = new Cleaner(whitelist);
+        Cleaner cleaner = new Cleaner(allowlist);
         Document clean = cleaner.clean(dirty);
         clean.outputSettings(outputSettings);
         return clean.body().html();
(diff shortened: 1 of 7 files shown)


drei01:
@moidurrahaman15 thanks for the approval. I'm not sure what your comment means?

src/main/java/org/jsoup/Jsoup.java
```diff
@@ -189,14 +189,14 @@ public static Document parse(URL url, int timeoutMillis) throws IOException {
 
      @param bodyHtml  input untrusted HTML (body fragment)
      @param baseUri   URL to resolve relative URLs against
-     @param whitelist white-list of permitted HTML elements
+     @param allowlist white-list of permitted HTML elements
```
nkiesel, line 192:
```suggestion
     @param allowlist list of allowed HTML elements
```

src/main/java/org/jsoup/Jsoup.java
```diff
@@ -206,48 +206,48 @@ public static String clean(String bodyHtml, String baseUri, Whitelist whitelist)
      tags and attributes.
 
      @param bodyHtml  input untrusted HTML (body fragment)
-     @param whitelist white-list of permitted HTML elements
+     @param allowlist white-list of permitted HTML elements
```
nkiesel, line 209:
```suggestion
     @param allowlist list of allowed HTML elements
```

src/main/java/org/jsoup/Jsoup.java
```diff
@@ -206,48 +206,48 @@ public static String clean(String bodyHtml, String baseUri, Whitelist whitelist)
...
      @see Cleaner#clean(Document)
      */
-    public static String clean(String bodyHtml, Whitelist whitelist) {
-        return clean(bodyHtml, "", whitelist);
+    public static String clean(String bodyHtml, Allowlist allowlist) {
+        return clean(bodyHtml, "", allowlist);
     }
 
     /**
      * Get safe HTML from untrusted input HTML, by parsing input HTML and filtering it through a white-list of
```
nkiesel, line 219:
```suggestion
     * Get safe HTML from untrusted input HTML, by parsing input HTML and filtering it through a list of
```

src/main/java/org/jsoup/Jsoup.java
```diff
@@ -206,48 +206,48 @@ public static String clean(String bodyHtml, String baseUri, Whitelist whitelist)
...
      * permitted tags and attributes.
```
nkiesel, line 220:
```suggestion
     * allowed tags and attributes.
```

src/main/java/org/jsoup/Jsoup.java
```diff
@@ -206,48 +206,48 @@ public static String clean(String bodyHtml, String baseUri, Whitelist whitelist)
...
      * <p>The HTML is treated as a body fragment; it's expected the cleaned HTML will be used within the body of an
      * existing document. If you want to clean full documents, use {@link Cleaner#clean(Document)} instead, and add
-     * structural tags (<code>html, head, body</code> etc) to the whitelist.
+     * structural tags (<code>html, head, body</code> etc) to the allowlist.
      *
      * @param bodyHtml input untrusted HTML (body fragment)
      * @param baseUri URL to resolve relative URLs against
-     * @param whitelist white-list of permitted HTML elements
+     * @param allowlist white-list of permitted HTML elements
```
nkiesel, line 227:
```suggestion
     @param allowlist list of allowed HTML elements
```

src/main/java/org/jsoup/safety/Cleaner.java
```diff
@@ -18,11 +18,11 @@
 
 
 /**
- The whitelist based HTML cleaner. Use to ensure that end-user provided HTML contains only the elements and attributes
+ The allowlist based HTML cleaner. Use to ensure that end-user provided HTML contains only the elements and attributes
  that you are expecting; no junk, and no cross-site scripting attacks!
  <p>
  The HTML cleaner parses the input as HTML and then runs it through a white-list, so the output HTML can only contain
```
nkiesel, line 24:
```suggestion
 The HTML cleaner parses the input as HTML and then runs it through an allow-list, so the output HTML can only contain
```

nkiesel:
Consider added a deprecated `Whitelist` back because otherwise semVer will demand that this becomes version 2.0.x.
Something like the following should suffice:
```java
package org.jsoup.safety;

/**
 * This class is deprecated and will be removed in version 2. Please change your code to use the class {@link Allowlist} instead.
 */
@Deprecated
public class Whitelist extends Allowlist {
}
```

drei01:
@jhy Do you agree that a deprecated `Whitelist` class should be present still? I'm happy to add it back if that's what's required to get this into a release.

src/main/java/org/jsoup/safety/Allowlist.java
```diff
@@ -50,50 +50,50 @@ If you need to allow more through (please be careful!), tweak a base whitelist w
...
- If you are going to extend a whitelist, please be very careful. Make sure you understand what attributes may lead to
+ If you are going to extend an allowlist, please be very careful. Make sure you understand what attributes may lead to
  XSS attack vectors. URL attributes are particularly vulnerable and require careful validation. See 
  http://ha.ckers.org/xss.html for some XSS attack examples.
  </p>
 
  @author Jonathan Hedley
  */
-public class Whitelist {
+public class Allowlist {
```
fprochazka, line 66: this is a big BC Break, changes like this should be done gradually with deprecations and a migration path

nkiesel: This is why I proposed to keep `Whitelist` as a deprecated alias for a couple of releases.

drei01: I've added the @deprecated Whitelist class as suggested.

fprochazka: IMHO this is not enough, you also have to have overloaded variants of methods where you've changed the signature

nkiesel:
Yes, unfortunately you are correct.  The only way forward I can see is to
add a full-fledged deprecated `Whitelist`.  We might be able to use
delegation to `Allowlist` so that the code is not completely duplicated
(e.g. which tags are part of "basic") . Something like the following
```java

@Deprecated
public class Whitelist extends Allowlist {

    private static Whitelist white(Allowlist allowlist) {
        Whitelist cloned = new Whitelist();
        cloned.tagNames = allowlist.tagNames;
        cloned.attributes = allowlist.attributes;
        cloned.enforcedAttributes = allowlist.enforcedAttributes;
        cloned.preserveRelativeLinks = allowlist.preserveRelativeLinks;
        return cloned;
    }

    private Allowlist allow() {
        Allowlist cloned = new Allowlist();
        cloned.tagNames = tagNames;
        cloned.attributes = attributes;
        cloned.enforcedAttributes = enforcedAttributes;
        cloned.preserveRelativeLinks = preserveRelativeLinks;
        return cloned;
    }

    public static Whitelist none() {
        return white(Allowlist.none());
    }

    public static Whitelist simpleText() {
        return white(Allowlist.simpleText());
    }

    public static Whitelist basic() {
        return white(Allowlist.basic());
    }

    public Whitelist addTags(String... tags) {
        return white(allow().addTags(tags));
    }
}
```

nkiesel:
I will push an alternative pull request that adds full backwards compatibility (with added test cases).

nkiesel:
Pull request #1423 now contains the same but with full backward compatibility

drei01:
Closing in favour of #1423

jhy:
Thanks - this is implemented now with #1464

PR #1422, opened 2020-08-15
Fix regression from commit a657ae

tilln:
The refactoring of the `Elements.forms()` method in [commit a657ae](https://github.com/jhy/jsoup/commit/a657ae0240b875a703a1e4909d56ad59997d362d#diff-4f0778cbabadac4729bae10ebec0c0aaR639) introduced a regression issue:
A `FormElement` in an `Elements` collection is not returned by `forms()` any more.
The `instanceof` operator equivalent [`Class.isInstance`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isInstance-java.lang.Object-) is incorrectly used.


tilln:
Already covered by #1403

krystiangorecki:
So which one is better?
I didnt't test it, but this one changes only one line of code.

PR #1423, opened 2020-08-19
Avoid the term whitelist

nkiesel:
This term was replaced with allowedList. See e.g. https://tools.ietf.org/id/draft-knodel-terminology-00.html
for a rationale.

### DIFF
--- a/src/test/java/org/jsoup/safety/CleanerDeprecatedTest.java
+++ b/src/test/java/org/jsoup/safety/CleanerDeprecatedTest.java
@@ -15,24 +15,24 @@ import static org.junit.jupiter.api.Assertions.*;
  Tests for the cleaner.
 
  @author Jonathan Hedley, jonathan@hedley.net */
-public class CleanerTest {
+public class CleanerDeprecatedTest {
     @Test public void simpleBehaviourTest() {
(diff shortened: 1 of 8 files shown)


drei01:
This looks good to me, thanks for picking up on my original request.

drei01:
Any updates on this PR?

jhy:
Thanks, will plan to get this in for 1.14.1.

src/test/java/org/jsoup/safety/CleanerDeprecatedTest.java
```diff
@@ -0,0 +1,331 @@
+package org.jsoup.safety;
```
jhy, line 1: Per other test comment, this should not be an entire redundant copy. Just the deprecated names.

jhy (changes requested):
Thanks for this - general approach is good but please see notes around preserving the file change history, and not duplicating the entire test files redundantly.

jhy:
I am happy to make the noted changes BTW, LMK.

jhy:
Thanks - rolled this mostly in to #1464 with a few minor changes (called it Safelist, mainly)

PR #1426, opened 2020-09-17
扩展图片src和a标签href的安全校验

hotlcc:
使其能支持多种类型的链接


jhy:
Thanks, but I don't think we want to add that to the defaults.

PR #1427, opened 2020-09-18
Add @SafeHtml java validator

### LINKED ISSUE #1382: Add @SafeHtml Jakarta Validator
Hibernate Validator currently has [@SafeHtml](https://docs.jboss.org/hibernate/stable/validator/api/org/hibernate/validator/constraints/SafeHtml.html), a Jakara Validator, to do validation of a string to ensure it's valid HTML using jsoup.

However, Hibernate Validator has decided to deprecate this validator as that project doesn't feel that it's the right place for it. Please see https://in.relation.to/2019/11/20/hibernate-validator-610-6018-released/ for more information.

I think jsoup is the perfect home @SafeHtml.

Copying @SafeHtml into jsoup would be done by:
* adding an Maven optional dependency on `jakarta.validation:jakarta.validation-api`
* copying a few classes for Hibernate Validator
* and adding a `META-INF` service loader file so the validator is loaded (see https://in.relation.to/2017/03/02/adding-custom-constraint-definitions-via-the-java-service-loader/ for details)

candrews:
This change does not add any new dependencies to jsoup.
If a bean validation implementation is present, then the new
@org.jsoup.constraints.SafeHtml constraints can be used.

This implementation was copied from Hibernate Validator 6.1.5.FINAL
SafeHtml was marked deprecated in Hibernate Validator following
CVE-2019-10219 see https://in.relation.to/2019/11/20/hibernate-validator-610-6018-released/

The root cause of the CVE was a misunderstanding of jsoup by Hibernate
Validator. By making the @SafeHtml constraint part of jsoup, that
concern is eliminated.

With this change, users have a logical source for @SafeHtml,
eliminating the need for each project to build and maintain @SafeHtml
itself, making for more secure and maintainable software.

Closes #1382


jhy:
Thanks, but I'm not keen on taking this in. Firstly, this looks like it's all copied directly from the Hibernate codebase which is licensed as LGPL and therefore not compatible with jsoup's MIT license. Secondly, I generally disagree with the validation approach used. The intended use of the jsoup cleaner is to *always* pass any input through the cleaner, and use the generated output. The `isValid` is meant as a UI indicator for end-users. The Hibernate approach (as I understand it, from memory) leads to this game of whack-a-mole and hence their deprecation of the code.

PR #1434, opened 2020-10-14
Added github actions to test the build after each push

HanSolo:
Added github actions to test the build after each push on Ubuntu 18.04 using JDK 8, 9, 11, 12, 13, 14 and 15


jhy:
Thanks for this - I merged a similar PR in #1458

PR #1440, opened 2020-10-26
Remove unused TokenQueue.consumeTagName

hannibal218bc:
This method is not used anywhere, and I guess it could easily be mistaken for `consumeElementSelector` so I'd suggest to remove it?


jhy:
This is a back-compat breaking change - not used internally but other users may still be hitting it. Maybe make it deprecated and we can pull in 1.15.

jhy:
Thanks - I marked a few unused methods in there as deprecated

5fc8403e5019db62f50057366a96174ef49a1761

Issue #1441, opened 2020-10-26
Unable to select element with a dot in its name


hannibal218bc:
As demonstrated here http://try.jsoup.org/~9rjkTKQu6z9ORnTAUY90s6p9k1w JSoup is currently unable to select elements with a dot in their elementName through `select(selector)`.
The dot is interpreted as "CSS class", so `node.red` is actually equivalent to `node[class=red]` and does **not** select the element `<node.red>` .

As outlined in #1055, for CSS IDs there is a workaround to use an attribute selector (`node[id=id.with.dot]`), but that won't work with element names. In that issue, it was also suggested to allow for escapes within the selector patterns, and I think that would be the only solution for element names.

Here's another fiddle with the proposed solution, instead of `node.red` you would write the selector as `node\.red`: http://try.jsoup.org/~AiYa0TBFS8HHzQ_y4QBfF3ZfASc

I'll shortly create a pull request with an implementation.

hannibal218bc:
So there's another PR #1443 with a different approach.

Issue #1055, opened 2018-05-02
Unable to select id has dot(.)


swarooppallapothu:
I am parsing some page. It has div like below.

`<div id="submit.add" class="pre current">Some text</div>`

I want to fetch all divs with "submit.add.pre". In jquery i am escaping dot(.) with **\\\\**. how i can do in JSOUP?

jhy:
Per the spec, ID's can't contain dots. So you can use the attribute selector, like this: `div[id=submit.add]`

E.g.: https://try.jsoup.org/~Rar9NhONcu3TbVc7WdvE90fAyAc

Backslash escapes would make sense, would welcome a PR.

mahozad:
The same problem here.

```kotlin
val element = """<h3 id="wear-phone-interactions-1.0.1">Wear-Phone-Interactions Version 1.0.1</h3>"""
val document = Jsoup.parse(element)

val first = document.select("#wear-phone-interactions-1.0.1").text()
val second = document.select("[id=wear-phone-interactions-1.0.1]").text()
```

The `first` one does not work. Had to use the `second` one.

jhy:
@mahozad not sure the point of this comment? I'm sure there are innumerable variations of invalid ID attributes.

mahozad:
Sorry, you're right. This is not a problem.

I was not aware that `#some.thing` selects an element with id `some` and class `thing` ([1], [2]).

  [1]: https://stackoverflow.com/a/1028260/8583692
  [2]: https://stackoverflow.com/a/12310128/8583692

PR #1442, opened 2020-10-26
Allow backslash-escaping within ElementSelector and CssIdentifier

hannibal218bc:
There has been previous discussion about element names or CSS tokens that can currently not been selected through JSoup's selector, because they contain JSoup Selector operators (like `#`, `.` , ...).

In #1055 it was suggested to use escape characters within the selector syntax.

This (simple) implementation modifies `TokenQueue` so `consumeElementSelector` and `consumeCSSIdentifier` will continue matching if the preceeding character is a backslash (`\`). Finally, all backslashes are removed from the result.

It is a simple approach, and will not allow escaping backslashes themselves - but there should never be a backslash in an element name or CSS identifier?

Here's a simple test case:
```java
	@Test public void testSelectorEscapes() {
		Document doc = Jsoup.parse("<root><e.dot>e with dot</e.dot><e class=\"dot\">e with class dot</e><f id=\"i.d\">f with i.d</f><x id=\"i\" class=\"d\">id i with class d</x></root>", "", Parser.xmlParser());

		assertEquals("e with class dot", 	doc.select("e.dot").text());
		assertEquals("e with dot", 			doc.select("e\\.dot").text());
		assertEquals("id i with class d", 	doc.select("#i.d").text());
		assertEquals("f with i.d", 			doc.select("#i\\.d").text());

	}
```

This PR would close issue #1441 .

### DIFF
--- a/src/main/java/org/jsoup/parser/TokenQueue.java
+++ b/src/main/java/org/jsoup/parser/TokenQueue.java
@@ -360,10 +360,10 @@ public class TokenQueue {
      */
     public String consumeElementSelector() {
         int start = pos;
-        while (!isEmpty() && (matchesWord() || matchesAny("*|","|", "_", "-")))
+        while (!isEmpty() && (matchesWord() || matchesAny("*|","|", "_", "-", "\\") || pos>0 && queue.charAt(pos-1)=='\\'))
             pos++;
         
-        return queue.substring(start, pos);
+        return queue.substring(start, pos).replace("\\", "");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


hannibal218bc:
Updated the PR to use `TokenQueue.unescape` to remove the escape characters, and added a flag to only do it when an escape char was detected in the loop.

jhy:
Hi Hannes,

Thanks for this, and sorry for the late reply. This update looks great. One suggestion:

> It is a simple approach, and will not allow escaping backslashes themselves - but there should never be a backslash in an element name or CSS identifier?

[Un]fortunately, HTML does indeed allow backslashes, both in tag names and in attribute names:
```html
<p\p attri\bute>Foo</p\p>
```
is valid and parses as-is (both in [jsoup ](https://try.jsoup.org/~uXLCCeAV7ZgGzMPtLKlSf4J2ISM)and in Chrome)

I think that we should enable the backslashes to be escaped.

LMK if you want to do that, or if I should.

Finally, it would be great if you could update CHANGES and most importantly, add a test-case for each case, to make sure it works and we never regress.

jhy:
(Somewhat amusingly, `<table(╯°□°)╯>Hello!</table(╯°□°)╯>` is also a valid tag name, although I wasn't able to create a selector for it, either with [jsoup](https://try.jsoup.org/~LCgRf2SNI9MiPG_euuIHaZr7BJY) or Chrome. And I don't think that's a requirement for this PR, but a nice i18n stretch goal :) )

PR #1443, opened 2020-10-27
Add Selectors :local-name, :name and :namespace-uri

hannibal218bc:
Another shot at #1441 by adding XPath-like Selector functions `:local-name()`, `:name()` and `:namespace-uri()`.

Here's a corresponding test case:
```java
	@Test public void testXMLXPathSelectors() {
		Document doc = Jsoup.parse("<root xmlns:xyz=\"https://www.eekhoorn.xyz/demo\" xmlns:abc=\"https://www.eekhoorn.xyz/demo\"><xyz:demo>demo xyz</xyz:demo><abc:demo>demo abc</abc:demo></root>", "", Parser.xmlParser());

		assertEquals("demo xyz demo abc", 	doc.select(":local-name(demo)").text());

		assertEquals("demo xyz", 	 	doc.select(":name(xyz:demo)").text());
		assertEquals("demo xyz demo abc", 	doc.select(":namespace-uri(https://www.eekhoorn.xyz/demo)").text());
	}
```

As it is with XPath, the matchers operate in a case-sensitive way, regardless of the document's ParserSettings.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -1620,4 +1620,29 @@ public class Element extends Node {
             && previousSibling() != null
             && !out.outline();
     }
+
+	public String namespacePrefix() {
+		final String tn = this.tagName();
+		final int p = tn.indexOf(':');
+		if (p<0) return null;
+		return tn.substring(0, p);
+	}
+
+	public String namespaceURI() {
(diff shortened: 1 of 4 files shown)


jhy:
Thanks - will close this now that Jsoup both supports XPath and (separately) escaped CSS selectors.

PR #1451, opened 2020-11-12
Fix NullpointerException in Element.isInlineable()

hannibal218bc:
Regression, introduced in 3a9bf46eeb36196d841c3fa05597e40a13f51f43 .

`parent()` will be `null` for freshly cloned, created or removed `Element`s.


hannibal218bc:
Oh, it's the identical issue to #1444

jhy:
Thanks for this. Same change was included in 77fcaf4dcdc8fd185e2f5d8eb6aa9c425a38bc9b when adding nullability annotations.

PR #1452, opened 2020-11-16
Limit reading CharBuffer.array() to CharBuffer.limit() to avoid additional NULL characters at the end of some inputs

hannibal218bc:
It is not documented explicitly, but `CharBuffer.array()` "Returns the char array that backs this buffer", so the buffer may have excess space at the end that was not used for the input.

(See also the explanation why https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6442955 is not a bug: 

> The Charset Buffer API is optimized for efficiency, not convenience. Buffer copies are avoided, but this means some buffers have extra elements between their limit and their capacity.

)

In a recent change to DataUtil the `array()` was passed without restricting its use to the length of `limit()`, hence adding NULL characters at the end.

Fixed by passing the value of `limit()` to the `CharArrayReader` that is actually reading from the `array()`.


jhy:
Thanks! Merged.

PR #1455, opened 2020-11-20
Use constants for the standard charsets (#1454)

### LINKED ISSUE #1454: Thread contention in OutputSettings()
Thread contention can happen in the constructor of `OutputSettings`.
image

It can be solved by replacing `Charset.forName("UTF8")` with the constant `StandardCharsets.UTF_8`.

jiri-pejchal:
Also solves thread contention in the constructor of OutputSettings.
Fixes #1454


jhy:
Thanks!

jhy:
I removed a few more with c0b275361ce8409f85de3573d89f6a617f0084d5

jhy:
However, using StandardCharsets breaks Android compat < API 19. So I moved to just define UTF8 as a static final, same as StandardCharsets but without the dependency.

PR #1464, opened 2020-12-15
Corrected uses of Whitelist to Safelist

jhy:
And introduced a (deprecated) compatibility shim.

Tested with japicmp binary and source compatibility report, and checked binary compat with some test utilities (local testing).

Many thanks to @drei01 (#1408), @nkiesel (#1423) and others for highlighting the issue and for PRs which this commit refines.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,10 @@ jsoup changelog
 
   * Change: updated the minimum Android API level from 8 to 10.
 
+  * Improvement: renamed the Whitelist class to Safelist, with the goal of more inclusive language. A shim is provided
+    for backwards compatibility (source and binary). This shim is marked as deprecated and will be removed in the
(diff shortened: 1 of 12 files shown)


jhy:
(It's annoying that git is losing the rename of Whitelist -> Safelist and so losing the history. Showed as a rename and re-add prior to commit, but not in history. Wonder if separate commits between the rename and the add would have kept clean.)

jhy:
I'm merging now, but would of course love for reviews and to hear any comments from everyone.

markaalvaro:
Do you have any idea when this will be released? I see I can get it in the snapshot, but was wondering if there'd be a stable release anytime soon that includes this change.

jhy:
> Do you have any idea when this will be released?

PR #1470, opened 2020-12-31
Attempt at a multi release jar

jhy:
This fails the animal-sniffer phase with "This feature requires ASM6". Supposedly, 1.19 supports this, but I can't get it to work.


jhy:
Done, finally, in #2025

PR #1476, opened 2021-01-16
Adds easy request session-state and corrects cookie domain and path rules

jhy:
This extends the Connection implementation to support (optional) sessions, which allow request defaults (timeout, proxy, etc) to be set once and then applied to all requests within that session. 

Cookies are re-implemented to correctly support path and domain filtering when used within a session. A default in-memory cookie store is used for the session, or a custom implementation (perhaps disk-persistent, or pre-set) can be used instead.

Forms submitted using the `FormElement#submit()` use the same session that was used to fetch the document and so pass cookies and other defaults appropriately.

The session is multi-thread safe and can execute multiple requests concurrently. If the user accidently tries to execute the same request object across multiple threads (vs calling `Connection#newRequest()`), that is detected cleanly and a clear exception is thrown (vs weird blowups in input stream reading, or forcing everything throw a synchronized bottleneck.

Examples:
```java
Connection session = Jsoup.newSession()
     .timeout(20 * 1000)
     .userAgent("FooBar 2000");

Document doc1 = session.newRequest()
     .url("https://jsoup.org/").data("ref", "example")
     .get();
Document doc2 = session.newRequest()
     .url("https://en.wikipedia.org/wiki/Main_Page")
     .get();
Connection con3 = session.newRequest();
```

### DIFF
--- a/src/main/java/org/jsoup/helper/CookieUtil.java
+++ b/src/main/java/org/jsoup/helper/CookieUtil.java
@@ -1,0 +1,88 @@
+package org.jsoup.helper;
+
+import org.jsoup.Connection;
+import org.jsoup.internal.StringUtil;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ Helper functions to support the Cookie Manager / Cookie Storage in HttpConnection.
+
+ @since 1.14.1 */
+class CookieUtil {
+    // cookie manager get() wants request headers but doesn't use them, so we just pass a dummy object here
+    private static final Map<String, List<String>> EmptyRequestHeaders = Collections.unmodifiableMap(new HashMap<>());
+    private static final String Sep = "; ";
+    private static final String CookieName = "Cookie";
+    private static final String Cookie2Name = "Cookie2";
+
+    /**
+     Pre-request, get any applicable headers out of the Request cookies and the Cookie Store, and add them to the request
+     headers. If the Cookie Store duplicates any Request cookies (same name and value), they will be discarded.
+     */
+    static void applyCookiesToRequest(HttpConnection.Request req, HttpURLConnection con) throws IOException {
+        // Request key/val cookies. LinkedHashSet used to preserve order, as cookie store will return most specific path first
+        Set<String> cookieSet = requestCookieSet(req);
+        Set<String> cookies2 = null;
+
+        // stored:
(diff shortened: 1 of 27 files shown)


src/main/java/org/jsoup/helper/CookieUtil.java
```diff
@@ -0,0 +1,88 @@
...
+        for (Map.Entry<String, String> cookie : req.cookies().entrySet()) {
+            set.add(cookie.getKey() + "=" + cookie.getValue());
+        }
+        return set;
+    }
+
+    static URI asUri(URL url) throws IOException {
+        try {
+            return url.toURI();
+        } catch (URISyntaxException e) {
```
jhy, line 79: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: While wrapping the caught exception into a custom one, information about the caught exception is being lost, including information about the stack trace of the exception.

**Fix**: If the caught exception object does not contain sensitive information, consider passing it as the &#34;rootCause&#34; or inner exception parameter to the constructor of the new exception before throwing the new exception. (Note that not all exception constructors support inner exceptions. Use a wrapper exception that supports inner exceptions.) 
[Learn more](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

jhy: Updated, thanks

src/main/java/org/jsoup/helper/CookieUtil.java
```diff
@@ -0,0 +1,90 @@
...
+        for (Map.Entry<String, String> cookie : req.cookies().entrySet()) {
+            set.add(cookie.getKey() + "=" + cookie.getValue());
+        }
+        return set;
+    }
+
+    static URI asUri(URL url) throws IOException {
+        try {
+            return url.toURI();
+        } catch (URISyntaxException e) {  // this would be a WTF because we construct the URL
```
jhy, line 79: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: While wrapping the caught exception into a custom one, information about the caught exception is being lost, including information about the stack trace of the exception.

**Fix**: If the caught exception object does not contain sensitive information, consider passing it as the &#34;rootCause&#34; or inner exception parameter to the constructor of the new exception before throwing the new exception. (Note that not all exception constructors support inner exceptions. Use a wrapper exception that supports inner exceptions.) 
[Learn more](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

jhy:
(Reminder, need to add details to the changelog, particularly a change note for those who may have implemented Connection directly, vs extending HttpConnection)

src/main/java/org/jsoup/helper/CookieUtil.java
jhy: Line 81 calls initCause(e), setting the root cause. It's just not in the constructor...

PR #1479, opened 2021-01-24
Minor performance improvement to Selector.select()

ehrmann:


src/main/java/org/jsoup/select/Selector.java
```diff
@@ -117,20 +117,19 @@ public static Elements select(String query, Iterable<Element> roots) {
         Validate.notEmpty(query);
         Validate.notNull(roots);
         Evaluator evaluator = QueryParser.parse(query);
-        ArrayList<Element> elements = new ArrayList<>();
+        Elements elements = new Elements();
```
ehrmann, line 120: `Elements` is already an `ArrayList`. The only reason to copy construct below is so it effectively `trims()` the list.

src/main/java/org/jsoup/select/Selector.java
```diff
@@ -117,20 +117,19 @@ public static Elements select(String query, Iterable<Element> roots) {
...
         IdentityHashMap<Element, Boolean> seenElements = new IdentityHashMap<>();
         // dedupe elements by identity, not equality
 
         for (Element root : roots) {
             final Elements found = select(evaluator, root);
             for (Element el : found) {
-                if (!seenElements.containsKey(el)) {
+                if (seenElements.put(el, Boolean.TRUE) == null) {
```
ehrmann, line 127: Maybe this is just being clever, but it avoids the second call.

jhy:
Thanks, merged!

PR #1480, opened 2021-01-24
Add missing Selector.select() convenience method

ehrmann:
`Selector` doesn't have a `select()` flavor that takes an `Evaluator` and multiple roots. Add one.

### DIFF
--- a/src/main/java/org/jsoup/select/Selector.java
+++ b/src/main/java/org/jsoup/select/Selector.java
@@ -115,8 +115,19 @@ public class Selector {
      */
     public static Elements select(String query, Iterable<Element> roots) {
         Validate.notEmpty(query);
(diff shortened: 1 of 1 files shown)


jhy:

ehrmann:
Is is that you don't think it belongs in the API, or are you just doing a winter PR cleaning? I can't remember my specific use case any more, but it does look like a missing version for `select()`.

jhy:
Mostly just doing (summer!) PR cleaning, and I didn't see a clear use case for it over using a stream. But could revisit if you do have a use case that others would likely share.

ehrmann:
I looked back at what I was writing when I opened this PR, and I can't remember what I was trying to do. The only one I came up with is precomputing the evaluator an being able to operate on `Elements`.

PR #1482, opened 2021-01-28
resolve abnormal urls in compliance with rfc3986

morokosi:
Some webpages may have abnormal URLs i.e. there are more ".." segments in a relative path reference than there are hierarchical levels in the base URL. Currently such extra segments are not resolved and remain as is. However, the extra segments should be removed according to 
https://tools.ietf.org/html/rfc3986#section-5.4.2 .
This will fix relative URL resolution and add test cases from rfc3986.

### DIFF
--- a/src/main/java/org/jsoup/internal/StringUtil.java
+++ b/src/main/java/org/jsoup/internal/StringUtil.java
@@ -9,6 +9,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.Stack;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


jhy:
Cool, thanks!

jhy:
BTW @morokosi for your interest: the original pattern could cause Stack Overflow with a nasty URL with a ton of `/../` segments. 

I fixed it by changing the original `^/((\\.{1,2}/)+)` to `^/(?>(?>\\.\\.?/)+)`. The SOE was in the `{` quantifier matching and the `(` capture.

#2165

PR #1494, opened 2021-02-22
CodeQL analysis

offa:
Adds a CodeQL analysis as part of the CI build. In addition there's a weekly cron, to check with latest queries regularly.


jhy:
Cool, thanks!

PR #1499, opened 2021-03-09
Adds optional tracking of discarded tags and atttributes

### LINKED ISSUE #1360: Report invalid elements or attributes
The _Cleaner_ class today uses the protected methods _isSafeTag_ and _isSafeAttribute_ to determine whether a tag or attribute is valid according to the provided whitelist. Those methods are not directly available to the application using Jsoup. To provide better feedback to whoever provided a document, it would be interesting to determine easily which tags and attributes invalidate a specific document. Maybe, in addition to keeping count of removed elements and attributes, the cleaner could also keep and provide a list of those things if requested.

Sebberh:
Closes #1360


Sebberh:
Forgot to squash commits

PR #1505, opened 2021-03-25
Refactored multiple assertion tests into parameterized ones

eas5:
Problem:
A test method with many individual assertions stops being executed on the first failed assertion, which prevents the remaining ones' execution. The difference between these assertions lies in different arguments only.

Solution:
Parameterized tests make it possible to run a test multiple times with different arguments. Using JUnit's parameterized tests feature, all assertions are executed as individual tests. This way, we could make 10 original tests become 55 independent ones. In this refactoring, no original assertion parameter was changed.


jhy:
Thanks, but I don't really get impacted by the current setup. Happy to use this style for new tests.

Issue #1492, opened 2021-02-09
Attributes equals now is order sensitive


hmdebenque:
Hi,

while updating to latest JSoup version I got unit test failing because `Attributes` equals implementation changed.
I believe it is a bug as the javadoc says:
> Attributes are treated as a map

As a simple test:
```java
final Attributes attributesBarFoo = Jsoup.parseBodyFragment("<a style=\"bar\" class=\"foo\">").select("a").first().attributes();
final Attributes attributesFooBar = Jsoup.parseBodyFragment("<a class=\"foo\" style=\"bar\">").select("a").first().attributes();
assertEquals(attributesBarFoo, attributesFooBar);
```
results in
```
org.opentest4j.AssertionFailedError: 
Expected : style="bar" class="foo"
Actual   : class="foo" style="bar"
```
I believe it was introduced in version `1.11.1` with commit ea1fb65e9ff8eee82c4e379dc3236d09a5ab02e1

Also, thanks for your great work on this amazing library.

LIKP0:
Hi, we are a student group and we would like to take a crack at this. Can't guarantee that we'll be able to complete it with high enough quality but we'll like to try.

LIKP0:
Hi, we are a student group and we would like to take a crack at this. Can't guarantee that we'll be able to complete it with high enough quality but we'll like to try.

LIKP0:
An early PR has been closed by us for adding more details (#1508). We will make a new PR soon.

zoey-shan:
Hi, I'm going to fix this issue by calling the `equals()` function for map directly to compare two `Attributes` since `Attributes`  can be changed to map by `dataset()` in `Attributes.java`

jhy:
Sorry for the late update! I've made the equals() method order insensitive.

Thanks for your feedback and for using jsoup!

jhy:
> Hi, I'm going to fix this issue by calling the `equals()` function for map directly to compare two `Attributes` since `Attributes` can be changed to map by `dataset()` in `Attributes.java`

That's a good idea, but dataset() is a filtered list of "data-" prefixed attributes.

PR #1508, opened 2021-04-13
Fix jsoup issue #1492

lexamxu:
Implement order-insensitive comparison with two attributes


lexamxu:
Please refer to the latest commit "enhanced fix jsoup issue#1492"

PR #1509, opened 2021-04-13
Add ability to set custom HostnameVerifier

uarlouski:
Hi @jhy, could you please review?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -42,6 +42,8 @@ jsoup changelog
 
   * Bugfix [Fuzz]: Speed improvement when the stack was thousands of items deep, and non-matching close tags sent.
     <https://github.com/jhy/jsoup/issues/1596>
+  
+  * Improvement: Add ability to set custom HostnameVerifier to verify hostname during handshake
 
 *** Release 1.14.1 [2021-Jul-10]
   * Change: updated the minimum supported Java version from Java 7 to Java 8.
--- a/src/main/java/org/jsoup/Connection.java
+++ b/src/main/java/org/jsoup/Connection.java
@@ -4,6 +4,7 @@ import org.jsoup.nodes.Document;
 import org.jsoup.parser.Parser;
 
 import javax.annotation.Nullable;
+import javax.net.ssl.HostnameVerifier;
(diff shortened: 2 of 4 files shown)


uarlouski:
Hi @jhy, could you please review?

giejay:
@jhy I'm also in need of setting a custom hostname verifier. Could you merge this request?

jhy:
Hi -- commenters, please don't add more ongoing pings to a PR or issue. It just creates clutter. If you'd like to show your interest, just vote with a thumbs-up (as an issue reaction).

I don't think that I'm going to merge this. When we have had code paths in jsoup previously that allowed SSL verification to be disabled, users on many platforms would get security audit warnings due to those code paths (even if they weren't called by the developer). Particularly, Android is very strict about this (rightly so, IMV), but there were other scanners that would throw warnings also.

And regardless of the warnings, I feel that it's a security footgun waiting to be used.

If you really have a need to disable hostname validation, why not use [#setDefaultHostnameVerifier?](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/HttpsURLConnection.html#setDefaultHostnameVerifier(javax.net.ssl.HostnameVerifier)) You could inspect the incoming hostname if you wanted to only apply it to some requests.

uarlouski:
Hi @jhy , thanks for your response

* I don't see any mentions about disabling SSL verifications in my pull request, HostnameVerifier is a part of java public api and anyone feel free to implement this.
* setDefaultHostnameVerifier also feels like footgun since it's applied to all places where the HttpsURLConnection is used
* I would expect that the API would be designed in the way where the users are not exposed to internals of the library that they use and interact with the library only through provided abstractions

Thanks :)

jhy:
@uarlouski the hostname verifier is used as part of the SSL evaluation, and Android scans for its use for applications disabling verification. Hence my points. I agree with you on the concept of the design of the API, but don't have a workable suggestion that won't trip the scanning of Android / Google Play.

draker94:
Hi, @jhy 

Any updates regarding this issue?
We also have problems with SSL verification on our project.

Thank you

Issue #1437, opened 2020-10-23
org.jsoup.nodes.Element.wholeText() not correctly translating <br> ?


JupiterRider:
`		String s = Jsoup.parse("Hello<br>World").wholeText();`
		`System.out.println(s);`

prints `HelloWorld`. 

But shouldn't it be
`Hello`
`World` ?

There is no line separator.

![firefox](https://user-images.githubusercontent.com/60042618/97023682-24944680-1556-11eb-9210-c8a93439d9e1.PNG)

![ide](https://user-images.githubusercontent.com/60042618/97023712-2eb64500-1556-11eb-84eb-d98cea55b974.PNG)

suarez12138:
Maybe it can be solved by adding detection of special tags like `<br>` and convert them manually in `wholeText()`. 

```JAVA
public void head(Node node, int depth) {
                if (node instanceof TextNode) {
                    TextNode textNode = (TextNode) node;
                    accum.append(textNode.getWholeText());
                }else if (node.nodeName().equals("br")){
                    accum.append("\n");
                }
            }
```

But whether there is a need to do it? And deal with more special tags like `<hr>` or just `<br>` ?

jhy:
Thanks this was fixed in #1636

PR #1516, opened 2021-04-16
Support html tags when getting text

suarez12138:
`<br>` and `<hr>` are not only semantic, they are also used like in #1437 . But should we deal with more special html tags?


jhy:
Thanks - similar was implemented in #1636

PR #1518, opened 2021-04-20
Fix issue #1492

lexamxu:
Fix issue #1492


jhy:
Thanks, and sorry for the late reply. In the future, please use a descriptive issue name (not just the bug number; who remembers those?) and link to the issue.

On revew -- I won't merge this -- the commit includes a lot of formatting changes which make the diff hard to read. It has odd comments like 'CS304' in it which don't add value. It adds a unnecessary new public class (which means a new contract to support) that is only used in this one spot. And the equals algorithm is now n^2. 

I think it would be simpler to use the entrySet equals method.

Issue #1490, opened 2021-02-08
Parsed head elements end up in body


waltkeh:
In this URL: https://www2.deloitte.com/us/en/insights/industry/power-and-utilities/future-of-energy-us-energy-transition.html

there is a valid HTML <head> element. When I analyse it with jsoup like:
jsoupDoc = Jsoup.parse( ... );
jsoupDoc.head();

the resulting jsoup head element is empty, but all the elements inside the HTML<head> are in the jsoup body element. This can be easily verified with the parsed output of TryJsoup:

 <head> 
 </head>
<body ng-app="appModule" ng-controller="appController">
  <section></section> 
  <title>Envisioning a sustainable energy future | Deloitte Insights</title> 
  <link rel="canonical" href="https://www2.deloitte.com/us/en/insights/industry/power-and-utilities/future-of-energy-us-energy-transition.html"> 
  <meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1"> 
  <meta http-equiv="content-type" content="text/html; charset=utf-8"> 
  <meta name="viewport" content="width=device-width, initial-scale=1.0"> 
  <meta name="robots" content="index, follow"> 
....

matjung:
In my home office, your observation happens when the parsable content starts with UTF BOM.

lexamxu:
You can use Jsoup.connect(url).get() instead. That will show the expected result

LIKP0:
You can use: 
1. Jsoup.parse(URL url ,int timeoutMillis) 
Use URL class instead of String a="https://www2.deloitte.com/us/en/insights/industry/power-and-utilities/future-of-energy-us-energy-transition.html"
2. Jsoup.connect(String url).get()

But it's a great idea to add a feature: 
1. Detect "html" is a network url but not the html file content when using Jsoup.parse(String html).
2. Use Jsoup.connect(String url).get() or other method to complete the process correctly.

I will make a PR later for this feature.

LIKP0:
An early PR has been closed by us for adding more details ([https://github.com/jhy/jsoup/pull/1508](https://github.com/jhy/jsoup/pull/1508)). We will make a new PR soon.

MonteCrystal:
[...]

PR #1519, opened 2021-04-21
Fix issue #1490

LIKP0:
Fix issue #1490


jhy:
Why are there so many duplicate PRs - are you all in the same class or something? Please be mindful if so to not create duplicate PRs, particularly for design changes like this. Closing, same reason as #1525 

Also - happy to take PRs but please make them clean (not multiple commits), and give a meaningful title.

Issue #1411, opened 2020-07-10
Element factory method


kenyons:
This is a feature request for a more concise way of creating Elements objects directly from an HTML String.  Right now, especially in unit tests, we have lots of code that looks like this:

`Element e = new Element("div").html("<span>Some stuff</span><span>Second part</span>");`

It would be very nice to have something more like this:

`Element e = Element.of("<div><span>Some stuff</span><span>Second part</span></div>")`

Or alternatively have a parser method like

`Element e = JSoup.parseElement("<div></div>");`

This is not only more concise, it would allow easier dynamic creation of mock elements for testing from a data file or list of HTML strings.

jhy:
Hi -- this makes sense and I think is a good idea. 

One issue with them being static methods (Element.of or Jsoup.parseElement) is that HTML parse settings cannot be applied or inherited from the current doc.

Maybe we can have it like `document.parseElement(html)`, like `document.createElement`.

Maybe both that and `Element.of(html)`

xiaoxianma:
To be consistent with Jsoup.parse, we may want a function named `Element.parse(String html)` instead of `Element.of`

xiaoxianma:
@jhy would you help review my pr? Appreciate!

jhy:
@xiaoxianma thanks - please see my comments on #1521. Similar thoughts on this.

jhy:
On further review I don't currently plan to implement this. I like the ergonomics, but because it doesn't allow any parser configuration (HTML vs XML, case sensitivity, tag set, etc) it is too limited.

Happy to take a look at other ideas.

PR #1521, opened 2021-04-24
Implement static method Element.of(html)

suarez12138:
I implement `Element.of(html)` mentioned in #1411 by using method `element.html()`, while using a `<div>` to be the top element can succeed in most situation, but can not hold cases like input tag `<html>`. I will glad to work more on this and welcome your code reviews.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -1628,9 +1628,11 @@ public class Element extends Node {
      * @see #html(String)
      */
     static Element of(final String html){
-        String head = html.trim().substring(1,5);
-        if (head.equals("html") || head.equals("head"))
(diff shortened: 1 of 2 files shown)


jhy:
Thanks, it's interesting. Sorry for the late reply! The main thing that I think is missing is that it's a static method and doesn't allow any of the configured HTML parser settings to flow (e.g case sensitivity, or XML vs HTML parsing, etc). So it could only be useful in the context of the default parser, which is frequently not what people actually want.

My suggestion would be to do something like making it a method of Document. Like `Element el = document.elementOf("<p>")...`.

And grab the parser from Document and use the parse fragment methods.

Also I think most of the validation, and particularly the string validation, should be removed. It seems both an artificial constraint, and doing string checks vs validating the parser output seems risky / unneccesary.

jhy:
Will close for now per my previous notes.

Issue #1514, opened 2021-04-16
replaceAll on Elements does not work


EugenMayer:
While this wont work:
```java
document.select("foo").replaceAll(element -> {
  return new Element("replacement");
});
```

this will work without any issues
```java
document.select("foo").forEach(element -> {
  element.replaceWith(new Element("replacement"));
});
```

If that makes any difference, in my concrete code i use 

```java
document.select(new Evaluator.Tag("pc:foo")).replaceAll(..);
```

But as far as i understand the implementation, that should not really matter

EugenMayer:
Just happened to notice that `replaceAll` is not at all implemented but inherited from `ArrayList` on `Elements` - seems somewhat very surprising offering a custom collection implementation which offers methods that will actually not lead to expected result.

I understand that this issue is then somewhat a feature-request/grey area bug but maybe needs at least to be addressed by: 
 - hiding non supported operations like this one
 - implementing the method

mct10:
Hi, I tried to implement this basically by overriding <code>replaceAll</code>. I think <code>node.replaceWith()</code> must be called, so as to change the element in the document. Simply calling <code>replaceAll</code> merely modifies the elements returned by <code>select</code>, leaving document itself unchanged. So, I call <code>replaceWith</code> in the new <code>replaceAll</code> and I think it works.

jhy:
Thanks - I implemented this in #2017 and the similar methods (like set, clear, replaceAll etc) to also update the backing DOM, as well as the Elements list.

PR #1522, opened 2021-04-24
Implement Elements.replaceAll

mct10:
**Summary**
I implement <code>replaceAll</code> mentioned in issue #1514 , by overriding the <code>replaceAll</code> in class <code>Elements</code>.
Also, I add a test to make sure the replacement happens as expected.


jhy:
Thanks - I implemented this with #2017, and added support for the other relevant methods (like set, remove, clear etc) to also update the DOM.

PR #1523, opened 2021-04-25
Fix #1492:Attributes equals now is order sensitive

### LINKED ISSUE #1492: Attributes equals now is order sensitive

zoey-shan:
Since Attributes are implemented by map, so the comparison can be realized by calling `equals()` for map in the function `equals()` in `Attributes.java`

### DIFF
--- a/src/main/java/org/jsoup/nodes/Attributes.java
+++ b/src/main/java/org/jsoup/nodes/Attributes.java
@@ -386,9 +386,7 @@ public class Attributes implements Iterable<Attribute>, Cloneable {
 
         Attributes that = (Attributes) o;
 
-        if (size != that.size) return false;
-        if (!Arrays.equals(keys, that.keys)) return false;
-        return Arrays.equals(vals, that.vals);
+        return dataset().equals(that.dataset());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


jhy:
Thanks and sorry for the late reply.

This implementation doesn't work, because dataset() is a filtered list of attributes that start with "data-". 

You're basically making equals() return true for every attribute set that does not include data- attributes.

Generally it's a good idea when writing test cases to include negative cases (in this case, things that should not be equal) to catch that kind of error.

Issue #1474, opened 2021-01-09
Empty attribute names produce invalid XML output syntax


jhy:
As noted in this [StackOverflow question](https://stackoverflow.com/q/65633287/153184), input with an empty attribute name produces invalid XML output when using the XML syntax. (I believe it's valid but meaningless HTML..)

**Input**
```
<div ="">
```

[produces](https://try.jsoup.org/~d9Ak3YlzQMWLcWkuIc49e0_8N_4):

**XML**
```xml
<div =""=""></div>
```

**HTML**
```html
<div =""></div>
```

Need to inspect the nodes directly for XML as I'm not sure what is happening there. The HTML at least looks valid.

The expected outcome of the XML output syntax is that an XML parser can read it - so probably we need to apply the same validations as in the W3CDom convertor. Specifically, the tag name and attribute validators.

For HTML it seems syntactically valid, but maybe a better idea that we just drop attributes with no key and no value -- what could be the point of retaining them? If they have a value though, should we keep an empty name? Happy to hear from people on this.

LIKP0:
Hello! I will try to fix it but could you give me some example codes to reproduce the bug?

LIKP0:
Bug reproduce example:
```
public static void main(String[] args) throws IOException {
        String ht = "<div someattribute=\"somevalue\"=\"\"></div>";
        Document doc = Jsoup.parse(ht);
        doc.outputSettings().syntax(Document.OutputSettings.Syntax.xml).charset(StandardCharsets.UTF_8);//Document.java 373
        String html = doc.html();
        System.out.print(html);
}
```

cqn2219076254:
I find that the problem is with the parser. For an input like <div ="">, the parser does not treat this as an attribute with no key and no value. Instead, it treats " ="" "as a whole as the key of attribute.

![图片](https://user-images.githubusercontent.com/71445886/115107272-64d00100-9f9c-11eb-83cd-6ee893184035.png)

XML automatically adds " ="" "  for attributes that do not have val, while HTML does not.

For example, if we input

```
<div a>
```

output of xml

```
<div a="">
```

output of html

```
<div a>
```

We can simply remove the XML auto-add ="" feature to make the results look correct. 

Or we need to rewrite parse so that it can parse attributes with no key and no val. But I think this will result in incorrect output like "< div >", because the program will not print an attribute with no key and no val, so we also need to make the program able to distinguish between "< div >" and "< div ="" >", which may be a big task.

[...]

PR #1524, opened 2021-04-25
Change how empty attribute names are handled

cqn2219076254:
When parsing <div ="">, the parser no longer treats " ="" " whole as the key of the attribute, but "","" as the key and val.
" ="" " will be retained if we use HTML, but not if we use XML


jhy:
Thanks -- wouldn't it be better to just drop attributes if they have no name? In the parse, for both HTML and XML. Unless there's a use case for preserving unnamed attributes?

jhy:
Closing, as corrected by #1474.

PR #1525, opened 2021-04-25
fix issue#1490

MonteCrystal:
Issue#1490 was because a url is passed while Jsoup.parse(String html) only accepts html. Created a check for wrongly passing url into jsoup.parse() where an html should be passed. If it is detected as url then handle it as url. Increased robustness.


jhy:
Thanks, but this is not something I want. Methods should not change behavior based on stringly typed parameters.

PR #1526, opened 2021-04-25
implement method replaceAll, instead of hiding non supported operations

Paramecium0:
#1514 When users want to replace all tags with other tags, they may use replaceAll(). But this will not work at all because the original method replaceAll() is inherited from `ArrayList` on `Elements`. To solve this issue, both implementing the method replaceAll() and hiding non-supported operations work. By using for loop and lambda, replaceAll() is implemented.


jhy:
Thanks - I implemented similarly and the other surrounding methods in #2017

PR #1527, opened 2021-04-25
Fix issue#1446

RyderCRD:
Fix issue#1446. HTML class names are case-sensitive, while our CSS selectors are case-insensitive. So I added a boolean variable to make the selection for Class can be customized by the user whether caseSensitive. This function is currently only for Class objects, if needed I can also add this option for other objects. I am a student taking a software engineering course and trying to fix Jsoup issues is part of my homework. My code may not be well written, thank you for your understanding, and welcome for any suggestions.


jhy:

PR #1529, opened 2021-04-26
Empty attribute name produces invalid XML output with using xmlParser

Ruefors:
Fix issue #1474
In my opinion, since the value of attribute name is empty, this result of parsing seem more reasonable

```
     <div></div>
```
Added two new variables hasIllegalAttributeName and illegalPendingAttributeName used to record the current key values are reasonable and while building the XML dom tree the illegal key value will change to empty

### DIFF
--- a/src/main/java/org/jsoup/parser/Token.java
+++ b/src/main/java/org/jsoup/parser/Token.java
@@ -71,26 +71,35 @@ abstract class Token {
             return forceQuirks;
         }
     }
-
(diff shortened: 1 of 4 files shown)


jhy:
Thanks - agree that we can just drop attributes if they don't have a name. But I don't see that we need these extra state values in the token , and fear that they will just lead to extra complexity. Seems simpler just to check and  drop on the emit / finalize stage.

Issue #1385, opened 2020-05-27
Incorrect detection of attribute `title` and `Element.text()`


mbamburova:
When parsing `title` and `text` from:

```
<a href="/code.php?set=HCPCS&c=A6454" title="Self-adher band w>=3" <5"/yd">A6454</a>
```

Attribute `title` is detected as **Self-adher band w>=3** instead of **Self-adher band w>=3" <5"/yd** 

and `text` is detected as **<5"/yd">A6454** instead of **A6454**. 
In previous version it worked fine.

For more such data:

```
<tr class="code-row" style='border-bottom:1px solid #f3f1f5;'>
   <td colspan='3' style='text-align:center;'><a href="/code.php?set=HCPCS&c=A6412" title="Occlusive eye patch">A6412</a></td>
   <td>&nbsp;</td>
   <td><a href="/code.php?set=HCPCS&c=A6412" title="A6412 Occlusive eye patch">Occlusive eye patch</a></td>
</tr>
<tr class="code-row" style='border-bottom:1px solid #f3f1f5;'>
   <td colspan='3' style='text-align:center;'><a href="/code.php?set=HCPCS&c=A6413" title="Adhesive bandage, first-aid">A6413</a></td>
   <td>&nbsp;</td>
   <td><a href="/code.php?set=HCPCS&c=A6413" title="A6413 Adhesive bandage, first-aid">Adhesive bandage, first-aid</a></td>
</tr>
<tr class="code-row" style='border-bottom:1px solid #f3f1f5;'>
   <td colspan='3' style='text-align:center;'><a href="/code.php?set=HCPCS&c=A6441" title="Pad band w>=3" <5"/yd">A6441</a></td>
   <td>&nbsp;</td>
   <td><a href="/code.php?set=HCPCS&c=A6441" title="A6441 Pad band w>=3" <5"/yd">Pad band w>=3" <5"/yd</a></td>
</tr>
<tr class="code-row" style='border-bottom:1px solid #f3f1f5;'>
   <td colspan='3' style='text-align:center;'><a href="/code.php?set=HCPCS&c=A6442" title="Conform band n/s w<3"/yd">A6442</a></td>
   <td>&nbsp;</td>
   <td><a href="/code.php?set=HCPCS&c=A6442" title="A6442 Conform band n/s w<3"/yd">Conform band n/s w<3"/yd</a></td>
</tr>
<tr class="code-row" style='border-bottom:1px solid #f3f1f5;'>
[...]

PR #1533, opened 2021-05-09
Fix issue #1385

LIKP0:
[https://github.com/jhy/jsoup/issues/1385](https://github.com/jhy/jsoup/issues/1385)


jhy:
Thanks - this seems to conflate various change. Creating a method that sometimes loads a URL and sometimes parses a string is not something I want to add to jsoup - methods should be unambiguous as much as possible. The attribute change moves jsoup out of spec from HTML5, which again, I don't want to do.

Issue #986, opened 2017-11-29
match with quoted pattern


aditsu:
I was checking how Jsoup handles `:matches(regex)` and noticed that the parsing code is not quite robust enough. Here's a plausible example that fails:
```
final Document doc = Jsoup.parse("<div>1) foo</div>");
System.out.println(doc.select("div:matches(" + Pattern.quote("1)") + ")"));
```
This is because `Pattern.quote` doesn't escape with backslashes like Jsoup expects, but rather wraps the text between `\Q` and `\E`, with some special handling if the original text contained `\E`.

HcPlu:
well,I have tried this case. I found out a very interesting thing that
` final Document doc = Jsoup.parse("<div>* foo</div>\n<div>\\)</div>");
System.out.println(doc.select("div:matches(" + Pattern.quote("\\)") + ")"));
System.out.println(doc.select("div:matches(" + Pattern.quote("*" + ")"));
`
it works actually.
but if , I use `Pattern.quote(")")`,it failed.So is `Pattern.quote("(")` and so char like )

So I guess maybe there is lake of some char parse like ")","(" ?

aditsu:
@HcPlu Your last line is missing a character.
Anyway, the bug is that Jsoup doesn't recognize the `\Q … \E` escapes, and crashes on an unbalanced `)` when parsing `matches( … )`.
Your first pattern works because Jsoup sees the `)` as being escaped by the `\`.
Your second pattern works (after fixing it) because it has no `)`.

HcPlu:
still keep my view. The (fixed)second case worked with pattern.quote.If Jsoup doesn't recognize `\Q'''\E`,how?
Taking your example,it still fails when exclude the `pattern.quote`

HcPlu:
@aditsu

aditsu:
Jsoup doesn't recognize `\Q` and `\E`, but after determining the regex, it creates a `Pattern` object from it, which DOES recognize `\Q` and `\E`.
So in other words, Jsoup works in 2 steps:
1) identify the regex by finding the closing `)` of the `:matches(` selector
2) create a `Pattern` from it and use it to test elements

Both steps involve some regex parsing, step 1 at a basic level and step 2 does the full job. Step 2 recognizes `\Q` and `\E` but step 1 does not, and that's where the bug is.

SliverySky:
[...]

PR #1536, opened 2021-05-11
Support for match with quoted pattern

suarez12138:
For `Pattern.quote()`, it turns a regex into a raw value, which if matched directly as an argument to `:match()` could cause an error like the one below mentioned in #986 due to lack of deliberate escaping.
```java
final Document doc = Jsoup.parse("<div>1) foo</div>");
System.out.println(doc.select("div:matches(" + Pattern.quote("1)") + ")"));
```
Therefore supports detection of the `\Q` and `\E` that  `Pattern.quote()` generates when parsing.


jhy:
Neato! Thanks, merged

Issue #763, opened 2016-09-24
Impossible to specify output settings for cloned elements


Sharcoux:
Cloned elements don't have an owner Document, which makes me unable to set output settings for these elements, and that leads to some issues in my app. Try this:

```
public class Main {
    public static void main(String[] args) {
        Document doc = Jsoup.parse("<div></div>");
        doc.outputSettings(new Document.OutputSettings().prettyPrint(false));
        Element e = doc.body();
        Element copy = e.clone();
        System.out.println(e.outerHtml());//returns html not pretty printed
        System.out.println(copy.outerHtml());//returns html pretty printed
        System.out.println(copy.ownerDocument());//returns null
    }
}
```

Of course, here it doesn't seem important, but what if you have 2 div with "display:inline-block" side by side ? The pretty printed html would render with an extra space between the 2 divs which is not the expected behaviour.

Shouldn't the outputSettings being copied to the cloned element ? Or the owner document being cloned as well as the cloned element ? Or at least, shouldn't we be provided a function to set outputSetting directly to an Element ? The owner document could be generated at that moment if needed ?

mct10:
Hi, I want to have a try on solving this.

jhy:
My suggestion would be to clone the parent Owner Document and attach that to the clone instead.

jhy:
Thanks for the suggestion - fixed! Only took a few years to get to ;)

PR #1547, opened 2021-05-16
modify Element.clone() for issue #763

mct10:
**Summary**
I add a private instance <code>Document.OutputSettings</code> to record what output setting this element itself should follow. Also, I modify <code>NodeUtils.outputSettings</code>, which obtains the output setting of the node. It checks whether the node is element or not. If it is element, than check if it has an ownerDocument. If it does not have an ownerDocument, it will use the element's output setting. Or it will still follow the ownerDocument's output setting.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Document.java
+++ b/src/main/java/org/jsoup/nodes/Document.java
@@ -393,6 +393,17 @@ public class Document extends Element {
             }
         }
     }
+
+    /**
+     * Set the document's output settings.
+     * This is used to override the function in Element, to avoid misuse.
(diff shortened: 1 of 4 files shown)


mct10:
This is an attempt to solve #763

jhy:
Thanks -- but I do not want to introduce another field to every Element in the DOM, as it really adds up how much memory is used in larger documents. Particularly for a (relative) edge case need as this.

My suggestion for a fix to the original issue would be to clone the parent Owner Document and attach that to the clone instead.

Issue #1550, opened 2021-05-20
NBSP creates invalid tags


rockwotj:
First - thanks for this library - it's great!

We just ran into an error where we got a bit of HTML that looks like this (encoded as a java string for clarity).

```java
String html = "<a\u00A0href=\"http://example.com/test\">test</a>"
```

This causes the tag to be parsed as: `a\u00A0href="http:`. Should the parser be made more tolerant to this sort of error? I'm assuming it's checking explicitly for ascii space - should any unicode whitespace character be a delimiter?

zoey-shan:
hey, I'm going to fix this issue

jhy:
That will pull it out of the HTML5 spec that jsoup targets, so I don't see this as something we should implement.

https://infra.spec.whatwg.org/#ascii-whitespace

PR #1551, opened 2021-05-21
fix issue #1550

zoey-shan:
Add operation to change '\u00A0' to a blank space in TokeniserState.java and CharacterReader.java


jhy:
Thanks - but NBSP is not whitespace per the HTML5 spec, so this breaks compliance.

(Why are we getting so many duplicate PRs - are you all in the same class or something? Can you chat with each other and not dupe?)

PR #1553, opened 2021-05-21
Support different whitespace characters

Paramecium0:
#1550 Improve the parser. It can detect three kinds of types of space in Unicode (i.e., \u00A0, \u0020, \u3000) now.

### DIFF
--- a/src/main/java/org/jsoup/parser/CharacterReader.java
+++ b/src/main/java/org/jsoup/parser/CharacterReader.java
@@ -377,6 +377,8 @@ public final class CharacterReader {
                 case '/':
                 case '>':
                 case '<':
+                case '\u00A0':
+                case '\u3000':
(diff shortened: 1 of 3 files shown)


jhy:
Hi - these don't seem to match the HTML5 spec for whitespace, so I don't think it's useful to make this change.

A potentially useful improvement if you want to make it would be to update the reader to match the definition in the HTML5 spec - it's currently missing some like Form Feed etc.

ASCII whitespace is U+0009 TAB, U+000A LF, U+000C FF, U+000D CR, or U+0020 SPACE.
https://infra.spec.whatwg.org/#ascii-whitespace

But, I haven't seen much use of this in the wild so am not sure there's much real-world impact.

jhy:
Will close for now per my previous notes.

PR #1554, opened 2021-05-22
optimized SearchTxet for simple format

Ruefors:
Added a simple method to solve the problem of extra space in searchText，The extra Spaces are merged


jhy:
Thanks, but I'm not sure what problem this is trying to solve. And the split[] and trim seems an odd way to further normalize spaces in lieu of a simple regex `\s{2,}`.

Anyway, `div:contains(write test)` matches the div so I'm not following the purpose here.

PR #1555, opened 2021-05-22
merging HEAD into BODY while use white list

Ruefors:
replace
```
Document dirty = parseBodyFragment(bodyHtml, baseUri);
```
to
```
Document dirty = parse(bodyHtml, baseUri);
```
Because that's what causes the head and body to blend together.
Then consider whether the head and body are in the white list respectively.


jhy:
Thanks -- planning on implementing with #2284

PR #1581, opened 2021-07-12
Dependabot: Ignore major Jetty updates

offa:
Ignores major updates of Jetty (#1574).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,9 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      # Jetty 9.x needed for JDK8 compatibility; it still receives security updates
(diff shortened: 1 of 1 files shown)


jhy:
Thanks - think this is better than the dependabor command I gave (which does look to have worked) because it's actually visible in code. These side-channel approaches hidden in bug chats seem like a good way to lose track of things.

offa:
Seems it doesn't work as expected 🙈:  #1589

offa:
Looks like the name should be `groupId:artifactId`. I'm going to submit a PR to fix this up.

jhy:
OK thanks - I wasn't sure if it was the config or the command that let it through. Difficulty with these GitHub configs is that it's hard to test them outside of waiting on a new event.

PR #1617, opened 2021-08-24
Fix #1616 by removing version from the javax.annotation OSGi package imports

### LINKED ISSUE #1616: OSGi import of javax.annotation and javax.annotation.meta is broken in 1.14.2
In jsoup version 1.14.2, the OSGi import of the package javax.annotation is imported with a version >= 3.0 and < 4.0.

This makes the jsoup 1.14.2 bundle fail to load on [apache karaf](https://karaf.apache.org) which provides version 1.3.0 of the package (from the apache felix runtime).

Possible fixes:
 
1. Check if the import is actually sed at runtime, and remove the import of javax.annotation, if it isn't actually needed (earlier versions of jsoup does not have this import)
2. Remove the versioning of the import (the actual content of the javax.annotation package has AFAIK not changed since, like, forever)
3. Expand the version range on the javax.annotation import, from [3.0, 4) to [1.0, 4)

Not sure where the 3.0 version of the import comes from...?
I have googled, and think, maybe from this 2011 vintage, org.glassfish rebundling of javax.annotation? https://mvnrepository.com/artifact/org.glassfish/javax.annotation

The javax.annotation.meta package will probably also have to be handled in the same way?
From the MANIFEST.MF of jsoup 1.14.2:
```
Import-Package: javax.annotation;version="[3.0,4)",javax.annotation.meta
 ;version="[3.0,4)",javax.net.ssl,javax.xml.parsers,javax.xml.transform,
 javax.xml.transform.dom,javax.xml.transform.stream,org.jsoup;version="[
 1.14,2)",org.jsoup.helper;version="[1.14,2)",org.jsoup.internal;version
 ="[1.14,2)",org.jsoup.nodes;version="[1.14,2)",org.jsoup.parser;version
 ="[1.14,2)",org.jsoup.safety;version="[1.14,2)",org.jsoup.select;versio
 n="[1.14,2)",org.w3c.dom
```

steinarb:


jhy:
Super! Thanks for this.

I took a quick pass at adding a testcase (reading /META-INF/MANIFEST.MF and checking the contents) but couldn't quickly determine how to get an integration test to run after the manifest build stage.

steinarb:
> Super! Thanks for this.
> 
> I took a quick pass at adding a testcase (reading /META-INF/MANIFEST.MF and checking the contents) but couldn't quickly determine how to get an integration test to run after the manifest build stage.

FWIW An integration test, checking the load-ability of the OSGi bundle, would be easier to add if this was a multi-module maven project.

Note also: I discovered a little too late that this doesn't fix the issue with the javax.annotation.meta import, because that's only in the findbugs jar and not in javax.annotation as such.

My plan when I get home tonight is to check if javax.annotation.meta is only used in test code, and if so, remove the javax.annotation.meta import from MANIFEST.MF

PR #1625, opened 2021-08-31
New GH Actions caching option

offa:
`setup-java` supports caching now :+1::

https://github.blog/changelog/2021-08-30-github-actions-setup-java-now-supports-dependency-caching/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         # choosing to run a reduced set of LTS, current, and next, to balance coverage and execution time
-        java: [8, 11, 15, 16]
+        java: [8, 11, 16]
       fail-fast: false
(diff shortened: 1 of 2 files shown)


offa:
The update to `v2` requires a distribution. I've selected the one of the example, but no preference from my side.

jhy:
Neat, thanks!

jhy:
By changing the cache key from {os}-{java version} to {os}, this also moves to a common Maven cache per OS, vs per OS+Java version. Not sure if that would improve the hitrate or not (as on a POM change they will all run concurrently and rebuild, and some will fail to save the cache as the key is already used). But it's nice to be using a little less storage.

Verified that subsequent runs do re-use the cache correctly: 
```
Cache restored successfully
Cache restored from key: setup-java-macOS-maven-1baf059fd606507d7fbc4bccfa5c4a7e6c70d8a7e27a9397c1fa623a073ba6c0
```
https://github.com/jhy/jsoup/runs/3478680098

PR #1629, opened 2021-09-08
Native XPath support in jsoup

jhy:
This is a first draft of adding native XPath support to jsoup. It uses the Java provided XPath parser & evaluator, so should fully support all valid XPath expressions.

As a starting point, I've added the method:
```java
Elements el.selectXpath(String expression)
```

For example:
```java
String html = "<body><div><p>One</div><div><p>Two</div><div>Three</div>";

Elements els = doc.selectXpath("//div/p");
assertEquals(2, els.size());
assertEquals("One", els.get(0).text());
assertEquals("Two", els.get(1).text());
```

I'm planning on bringing this in as a *beta* feature in the next release (1.14.3 -- I'll add a few more tests first) and then hopefully finalizing it in a release shortly following that. I'm looking for all feedback on how to make this feature most useful, and feel jsoup native.

Particularly: 
- does the name `selectXpath` make sense? Had thought also about just `xpath()` or `selectX`)
- should we have a `selectXpathFirst` method, like `selectFirst`?
- this current implementation is namespace aware. From looking at common XPath questions, it seems that that namespaces can be a source of confusion and query difficulty. Should we make it default to namespace unaware, and add another method (?) `selectXpathNS` or similar with it on? Or best to leave as-is?
- Does this match what you were expecting when you heard "jsoup has xpath support now"? What's missing? (Please provide suggested API signatures and how you'd use them)
- Does it work for your data and queries? Can you break it?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,7 +3,7 @@ jsoup changelog
 *** Release 1.14.3 [PENDING]
   * Improvement: added native XPath support in Element#selectXpath(String)
     <https://github.com/jhy/jsoup/pull/1629>
-    
+
   * Improvement: added support in CharacterReader to track newlines, so that parse errors can be reported more
     intuitively.
     <https://github.com/jhy/jsoup/pull/1624>
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,9 @@
 jsoup changelog
 
 *** Release 1.14.3 [PENDING]
+  * Improvement: added native XPath support in Element#selectXpath(String)
+    <https://github.com/jhy/jsoup/pull/1629>
+    
   * Improvement: added support in CharacterReader to track newlines, so that parse errors can be reported more
     intuitively.
     <https://github.com/jhy/jsoup/pull/1624>
--- a/pom.xml
+++ b/pom.xml
@@ -345,6 +345,15 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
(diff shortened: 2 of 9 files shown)


manticore-projects:
Greetings!
First and most important: Thank you so much for your work and providing those great tools! Much appreciated.

Now some requested feedback from practice. Parsing an XHTML file in order to insert a TOC, I got everything working somehow, but it was not straight forward and a lot of trial'n error.

The main confusion is about the XPATH. With my limited knowledge, I would have expected this to work (it works 
```xsl
//xhtml:a[not(@href)]/@name
```

But in JSoup, I had to use
```xsl
//*[local-name()='a' and not(@href) and @name]
```

I get somehow, why I have to work around the namespaces. But I am totally lost, why this did not work
```xsl
//*[local-name()='a' and not(@href)]/@name
```

XSoup did not work at all because of the Name Space and looking for elements without the `href` attribute.

jhy:
@manticore-projects thanks for the feedback. Can you take a look at the current release (1.15.2, you can use [try.jsoup](https://try.jsoup.org/)) and see if your issues are fixed? Otherwise, could you open a new issue and include sample HTML / XML, and the xpath query. I have disabled namespace awareness in the xpath evaluator which will simplify things.

GitHubDaniel:
Works even better than xsoup. Thanks a lot. However the only thing i experienced was "Could not find function: lower-case". That's not a problem for me since [contains(attr, val) does ignore case in jsoup.

It's technically 2 errors which even out for me ;-)

Issue #1647, opened 2021-09-29
Allow attributes valid in html when converting from JSoup to W3C Document


jairamc:
Consider the following html document:

```html
<!DOCTYPE html>
<html>
<head></head>
<body style="color: red" " name">
  <p hành="1" hình="2">unicode attr names</p>
</body></html>
```

Using v1.14.2 and running the following code:

```java
    public static void main(String[] args) {
        String html = "<!DOCTYPE html><html><head></head><body style=\"color: red\" \" name\"><p hành=\"1\" hình=\"2\">unicode attr names</p></body></html>";
        org.jsoup.nodes.Document jsoupDoc;
        jsoupDoc = Jsoup.parse(html);
        Document w3Doc = W3CDom.convert(jsoupDoc);
        System.out.println(W3CDom.asString(w3Doc, W3CDom.OutputHtml()));
    }
```

Results in:

```html
<!DOCTYPE html SYSTEM "about:legacy-compat">
<html>
<head><META http-equiv="Content-Type" content="text/html; charset=UTF-8"></head>
<body name="" style="color: red">
  <p hnh="2">unicode attr names</p>
</body>
</html>
```

This is caused by [W3CDOM.java#L346](https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/helper/W3CDom.java#L346) hard-codes the syntax to `xml`. It can be easily fixed by checking the doctype of the output document and use that as the syntax.

jhy:
Hi -- this looks like a good change. I made a comment on the PR (#1648) on the implementation, just wanted to check you'd seen it.

jairamc:
Hi @jhy . My sincere apologies. I have seen your response. I'm at a company conference this week and hence have been struggling on time to work on your suggestions. I'll try and push an update this week, if not, next week for sure.

jhy:
Thanks! I have merged this with a couple tweaks. And certainly nothing to apologize for, I was just checking in.

PR #1648, opened 2021-09-29
Allow attributes valid in html when converting

jairamc:
When parsing and converting an html document, the "syntax" was hard-coded to xml. This PR checks the document type of the output document and uses that to determine which attributes are valid.

Changes for https://github.com/jhy/jsoup/issues/1647

### DIFF
--- a/src/main/java/org/jsoup/helper/W3CDom.java
+++ b/src/main/java/org/jsoup/helper/W3CDom.java
@@ -38,10 +38,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Stack;
-import java.util.regex.Pattern;
+import java.util.Locale;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jsoup/helper/W3CDom.java
```diff
@@ -342,8 +344,15 @@ public void tail(org.jsoup.nodes.Node source, int depth) {
         }
 
         private void copyAttributes(org.jsoup.nodes.Node source, Element el) {
+            final Syntax syntax;
```
jhy, line 347: I think it would be better to get the syntax from the Document's parser settings. That's what we use for every other syntax selection. And, documents don't always have doctypes. Just pass it in as an argument to this method (vs calling ownerDocument() every time, which has to walk up the parent() stack).

PR #1655, opened 2021-10-17
Expose maxPaddingWidth in OutputSettings keeping default as 30

### LINKED ISSUE #1653: Please expose maxPaddingWidth in two ways
Hi,

I work with the formatter-maven-plugin and ever since 1.14.1, we have been unable to upgrade as it seriously corrupted pretty formatted html.  After deep diving this, I finally found the cause.  So I think the change basically breaks pretty formatting in most cases.  Our example is rather simple, you cna find it here -> 

[test](https://github.com/revelc/formatter-maven-plugin/blob/main/src/test/java/net/revelc/code/formatter/html/HTMLFormatterTest.java}
[resource](https://github.com/revelc/formatter-maven-plugin/blob/main/src/test/resources/someFile.html)

We keep the hash of what the prior formatter was producing (namely 1.13.1) that is on our master now.  If it changes during an upgrade, it should be some improvement.  In this case its rather corrupted by lack of proper whitespace.  It further doesn't line break on the header any longer (not solved on this ticket).

After reviewing all the code changes and a lot of debuging, I have gotten to the StringUtil change [here](https://github.com/jhy/jsoup/commit/afd73606a90909444e1c443b555dae7b71e6a5a0#diff-597d6f91192e47c0d6c1eb2451e4f36c54d9f3d982ccb867f30246712384506f).  Specifically line 75 (width = Math.min(width, maxPaddingWidth);) or line 128 in your lastest master.

After looking at what formatted for us, it was clear it was 30 characters as that is set preventing proper nesting to occur.  So I cloned this repo and commented out that line then ran against that.  The result was more of what we expected.  A few tags were shortened where possible (but note that header line break from star slash comment didn't change).  This is far more acceptable to us given our user base and fact that jsoup is vulnerable I believe a couple times now.

This is clearly internal code of jsoup and not intended for external exposure based on the package naming but I feel we need this exposed with 2 options.  The first option, completely disable it so it works like it did before as that is rather critical as we are formatting source code to be committed to repos.  The second option would simply to allow some sizing.  So I think something like -1 to mean no limit and then configurable to whatever a user might want.

I don't might writing the code to expose that out but wanted to run this full diagnostic of the issue by your team first before attempting to do so.

hazendaz:
Fixes #1653 

In order to get my fork up and going, I updated github actions here.  I removed jdk 16, added jdk 18-ea, and moved to zulu distribution as temurin does not have 18-ea.

For this change, I added a changelog, added some clarification to 'memoised' padding, deprecated original padding call by having it call newer one with exposed maxPaddingWidth, skipped the min check with -1, and finally called the new one.  In an attempt to better understand 'memoised' logic, I actually added more tests than necessary to the existing one for clarity purposes on how that method behaves.

The overall change though tripped up the japicmp plugin.  It kept saying the method had been removed.  If I removed the deprecation markings I added, it was happy.  So I think that plugin only really works against deprecations from prior version.  I tried a few different ways to make that work and finally gave up so I just added the class to ignore for the time being.  I believe that is a bug in their plugin and that deprecation is important if anyone is otherwise externally using that method since it directs them on what to update.

One other thing, on my fork, github no longer auto turns on actions so I had to get that working so I hunted for some basic items to get it going.  I noticed that jetty-servlet was not updated.  That is odd given dependabot usage and the other module from jetty was updated so I bumped it up.

Obviously this has more than what I should have for a new pull request.  For purposes of review, please use this.  If you do need me to then rebase and only send up direct change portions, no problem, I can rebase as necessary.

Thanks,

Jeremy

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,13 +17,13 @@ jsoup changelog
     element, e.g. ancestor-or-self::*.
     <https://github.com/jhy/jsoup/issues/1652>
 
+  * Improvement: allow a maxPaddingWidth on the indent level in OutputSettings when pretty printing. This defaults to
+    30 to limit the indent level for very deeply nested elements, and may be disabled by setting to -1.
+    <https://github.com/jhy/jsoup/pull/1655>
+
   * Bugfix: boolean attribute names should be case-insensitive, but were not when the parser was configured to preserve
     case.
     <https://github.com/jhy/jsoup/issues/1656>
-
-  * Improvement: allow maxPaddingWidth in OutputSettings for pretty printing.  This will continue to default to 30
-    so very deeply nested nodes don't get insane padding amounts but now allows for user decision.  Using -1 will
-    result in original unlimited length.
 
 *** Release 1.14.3 [2021-Sep-30]
   * Improvement: added native XPath support in Element#selectXpath(String)
--- a/src/main/java/org/jsoup/internal/StringUtil.java
+++ b/src/main/java/org/jsoup/internal/StringUtil.java
@@ -114,29 +114,28 @@ public final class StringUtil {
     }
 
     /**
-     * Returns space padding (up to a max of 30).  Use {@link #padding(int, int)} if you need
-     * more than 30 spaces padded.
+     * Returns space padding (up to the default max of 30). Use {@link #padding(int, int)} to specify a different limit.
      * @param width amount of padding desired
      * @return string of spaces * width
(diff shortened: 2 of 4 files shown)


hazendaz:
Great!  Thanks!

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #1662, opened 2021-11-04
[actions] Remove jdk 16, add jdk 18-ea and move from temurin to zulu …

hazendaz:
…to pick up jdk 18-ea


jhy:
Thanks - I don't particularly want to run the EAs just yet -- when we did that for v17, we got a lot of build failures, particularly on the Mac builds. I think it's best to wait for it to stabilize first.

hazendaz:
18 is ramp down now so it should be stable but that makes sense.  I can re-raise this and just use the 17 release to replace obsolete 16.

jhy:
Thanks, sounds good. Can stay on Temurin too I guess.

PR #1664, opened 2021-11-04
[header] Support header comments so line breaks properly

hazendaz:
context: formatter-maven-plugin

Issue: Many files come with copyright headers in the files.  The html parser for pretty printing with jsoup treats this header comment incorrectly and ends up adding the html tag at end of --> on same line.  I'm not sure the fix I have here is appropriate but it did solve the issue.

As with before you can see this with this file [here](https://github.com/revelc/formatter-maven-plugin/blob/main/src/test/resources/someFile.html).

### DIFF
--- a/src/main/java/org/jsoup/nodes/Comment.java
+++ b/src/main/java/org/jsoup/nodes/Comment.java
@@ -37,12 +37,23 @@ public class Comment extends LeafNode {
     }
 
 	void outerHtmlHead(Appendable accum, int depth, Document.OutputSettings out) throws IOException {
-        if (out.prettyPrint() && ((siblingIndex() == 0 && parentNode instanceof Element && ((Element) parentNode).tag().formatAsBlock()) || (out.outline() )))
+        if (out.prettyPrint() && ((siblingIndex() == 0 && parentNode instanceof Element && ((Element) parentNode).tag().formatAsBlock()) || (out.outline() ))) {
             indent(accum, depth, out);
-        accum
+        }
+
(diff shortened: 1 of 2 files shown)


hazendaz:
For now looking for feedback only.  Depending on feedback will adjust tests as needed.

hazendaz:
This is ready for review.  Basic idea here is license header comments used in html pages.  The current jsoup will not line break after the comment thus the <html> block start is at end of the header comment which is not correct.  To avoid touching other comments, this addresses only when 'accum' is empty and thus start of the file processing when pretty print is turned on.  An existing toString test already runs full pretty print and was able to be adjusted to confirm this test while keeping all other existing tests intact.

hazendaz:
@jhy Could you look at this one?

hazendaz:
You can see this saved item Saved https://try.jsoup.org/~pWWgM1Tq8l3YM1WTm8Fftq5ziv4, look at the header as it buts up against the docttype.  It should line break which this one attempts to fix and does work.

hazendaz:
rebased, still not fixed until this commit

hazendaz:
Jsoup Header formatting issue

Attached is after running jsoup pretty format.  Typically license header will be at top of the file and it fails to properly account for that.  See the <!doctype> is in end of the comment to header.  This fixes it issue.

hazendaz:
@jhy Finally got back to properly confirm, this worked as expected.  This also seemed to fix another issue I had not noticed previously where I ended up with 2 license headers.  Anyway, thank you!

PR #1665, opened 2021-11-04
fix #1411 add Element parse method

### LINKED ISSUE #1411: Element factory method

xiaoxianma:
Add a static method parse to build an element from HTML string

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -86,6 +86,34 @@ public class Element extends Node {
         this(tag, baseUri, null);
     }
 
+    private Element() {
+        this("-");
+    }
+
+    /**
+     * Parse HTML String into an Element
+     * <p>
+     * Use examples:
+     * <ul>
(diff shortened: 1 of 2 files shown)


xiaoxianma:
@jhy Could u help review this PR? thanks!

xiaoxianma:
> @jhy Could u help review this PR? thanks!

Any update? Thanks

jhy:
Please don't keep pinging on these like that. It just creates more noise; the PR is still in the list. You can always reach out to me directly if need be.

I left a comment on the original discussion bug.

Issue #1670, opened 2021-11-11
Parsing SequeceInputStream leads to loss of part of html content


KirylBubovich:
When using `DataUtil::parseInputStream` with `SequenceInputStream`, the returned document contains only part of the html, and the other part is lost. This is due to the fact that we read more than the specified max bytes in `ConstrainableInputStream::readToByteBuffer`. The consequence of this is that in `DataUtil` the expression `input.read () == -1` returns `true` (although the input stream has not actually been read completely).

jhy:
Thanks!

PR #1671, opened 2021-11-11
Fix sequence input stream parsing

KirylBubovich:
* the problem with SequenceInputStream is that part of the html content is lost

#1670


KirylBubovich:
To understand the problem more, please run the `parseSequenceInputStream` test with no changes to `ConstrainableInputStream`. Part of the html content should be missing

jhy:
Thanks! Merged

PR #1674, opened 2021-11-20
Add Methods to Elements Class

yanok9812:
This PR adds Elements.before(Node), Elements.after(Node), Elements.append(Node), and Elements.prepend(Node), alone with the unit tests.

For Issue #953


jhy:
Sorry for the late reply. Thanks for your interest. I'm not going to move ahead with this. The nodes shouldn't be inserted via serializing and re-parting (`toString()`), that's very inefficient, and not consistent with the rest of jsoup. They should be cloned instead.

PR #1676, opened 2021-11-26
Add CIFuzz integration to GitHub Actions

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which jsoup already runs.

CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above).

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
I think this should be working now, could we try to retrigger the CI?

jhy:
Good stuff, thanks!

PR #1683, opened 2021-12-13
Change insertNode and popStackToClose from private to protected

magicprinc:
so they can be re-used in jsoup-json's JsonTreeBuilder without copy-paste
https://github.com/magicprinc/jsoup.json


jhy:
Merged, thanks

PR #1685, opened 2021-12-19
Adding a new example

damienhunter30:
Added 2 classes as a first time code contributor. Gathers data from Distro watch, the 7 day trend.


jhy:
Thanks -- this looks quite coupled to that specific site and I'm not sure demonstrates jsoup particularly differently to the other examples. I think it might be better as a blog post.

PR #1698, opened 2021-12-23
Add url fragment field to `Document` class

### LINKED ISSUE #1686: Document::location does not contain fragment identifier
Hi,

It is important for me to get the fragment id of a page from jsoup *Document*s.  
Is it possible to get the URL intact as passed to the *Jsoup::connect* method?

The original URL that I pass to *Jsoup::connect* method has a fragment identifier at its end.
But *Document::location* or *Document::baseUri* do not contain the id.

For example, `https://developer.android.com/jetpack/androidx/releases/annotation`**`#annotation-1.4.0-alpha01`**
becomes `https://developer.android.com/jetpack/androidx/releases/annotation`.

mahozad:
Fixes #1686

### DIFF
--- a/src/main/java/org/jsoup/helper/DataUtil.java
+++ b/src/main/java/org/jsoup/helper/DataUtil.java
@@ -296,6 +296,20 @@ public final class DataUtil {
         return StringUtil.releaseBuilder(mime);
     }
 
+    /**
+     * Returns the fragment identifier of the url.
+     *
+     * It is called reference in {@code java.net.URL} class ({@code URL::getRef}).
+     *
+     * @param url the url string
+     * @return the url fragment part or empty string if the url contains no fragment
+     */
+    static String getUrlFragment(String url) {
+        int index = url.indexOf('#');
+        if (index < 0) return "";
+        return url.substring(index + 1);
+    }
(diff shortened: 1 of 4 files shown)


jhy:
Thanks for the PR.

I'm not sure that / why this needs to be a new method. Why should it not just be in the Document.location() string? If the user wants to parse component of that as a URL (say the host or the path), they can create a URL object and allow the native methods to parse it. It seems redundant and incomplete to special case the fragment / ref / hash like this.

Similarly, in the Connection object, I think it'll be best to grab it from the URL object to persist, vs trying to parse it in Data Util.

mahozad:
I initially thought that for existing programs that rely on the Document.location() and update to its new version (jsoup new version), they may have assumed that the return string contains no fragment id. So, I thought changing this behavior may cause trouble (is it called forward compatibility?).

But now I think they can update their code with a simple substring.  
This means, **they have to update their code** if the new behavior is problematic for them.

Please change/fix the code as you wish.
My solution may not be the best one.

jhy:
Thanks, but won't merge as the fragment is already preserved, and I don't see a need to add this new method.

PR #1721, opened 2022-02-13
added junit-5 extension to handle network flaky tests

unshorn:
Intercepts network related exceptions to skip tests that can be flaky due to network errors. The extension can be applied to all tests by adding the runtime parameter `-Djunit.jupiter.extensions.autodetection.enabled=true`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -352,6 +352,14 @@
       <version>3.0.2</version>
       <scope>provided</scope>
     </dependency>
+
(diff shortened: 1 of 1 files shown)


jhy:
Thanks - I think we should just get rid of those tests though...

jhy:
(103d7864a6d54391a2a98c5027e03d5a777ebc71 removed the tests using network access)

PR #1727, opened 2022-02-27
Bring dependencies up to date

hazendaz:


jhy:
Thanks - caught these with the dependabot PRs

PR #1728, opened 2022-02-27
[modernize] Use already compiled charsets instead of recompiling them

hazendaz:
retains support or older usage of andriod as indicated in the source file.  Test end didn't matter in that regard as seen by other code.


jhy:
Thanks! Merged.

PR #1738, opened 2022-03-25
Feature/refactoringv1

pooja-anandani:
I have reviewed the project and refactored few of the things that can improve the code readability and quality.

The tasks performed are:

[1] Rename variable:
In file XmlTreeBuilder.java there was an element object named el, this naming convention is not meaningful and doesn't provides the clarity to the developer hence, I have changed the name and it's reference to new_standalone_element.

[2] Extract Method:
Sometimes, the logical implementation of the method is too long so, to improve it I extracted method where multiple conditional statements were there all of them were moved to addAttributes in DocumentType.java. Same is followed in DataUtil.java.

[3] Pull up method:
There were few child classes that were using the same method so, I pushed those method in parent class LeafNode as they were not having any specific implementation it can be seen in method outerHtmlTail.

[4] Extract Class:
The insufficent modularization is being solved there was a class OuterHtmlVisitor.java in Node that can be extracted to different class.

[5] Move Method:
lastCharIsWhitespace() this method was declared in TextNode.java and it was used only by Element.java, so that method was moved to Element.java


jhy:
Thanks for the feedback. Looks like mostly a matter of differing styles. It's a bulky commit and tough to review, including unwanted whitespace changes, dead code, etc. Not inclined to merge.

PR #1744, opened 2022-04-08
Use moditect for multi-release module-info

Meodinger:
Jackson was using [moditect](https://github.com/moditect/moditect) to generate multi-release module. 
I've tested on release 1.14.3.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
   <groupId>org.jsoup</groupId>
   <artifactId>jsoup</artifactId>
-  <version>1.14.3</version><!-- remember to update previous version below for japicmp -->
+  <version>1.14.3-module</version><!-- remember to update previous version below for japicmp -->
   <description>jsoup is a Java library for working with real-world HTML. It provides a very convenient API for fetching URLs and extracting and manipulating data, using the best of HTML5 DOM methods and CSS selectors. jsoup implements the WHATWG HTML5 specification, and parses HTML to the same DOM as modern browsers do.</description>
   <url>https://jsoup.org/</url>
   <inceptionYear>2009</inceptionYear>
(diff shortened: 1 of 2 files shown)


Meodinger:
Strange, github checks all failed because of moditect: Invocation of jdeps failed

Meodinger:
Tested on master, `jsr305` is not required in module-info because it's in the scope `provided` and it's not modular.

pom.xml
```diff
@@ -123,9 +131,6 @@
         <version>3.2.0</version>
         <configuration>
           <archive>
-            <manifestEntries>
-              <Automatic-Module-Name>org.jsoup</Automatic-Module-Name>
```
jhy, line 127: Where does the module name come from now - is it line 247 module artifact?

src/main/moditect/module-info.java
```diff
@@ -0,0 +1,9 @@
+module org.jsoup {
+    requires static java.xml;
+
+    exports org.jsoup;
```
jhy, line 4: Don't we need to export .helper as well? I am concerned that tests didn't fail for this, are there tests to ensure we are exporting the correct packages?

Meodinger: Actually, I'm not very clear with the package structure of Jsoup and not very sure who should be exported and who should not, so I just export `org.jsoup`. I'll add the necessary `exports` if you suggest some.

jhy:
Thanks all - I have module support with #2025.

It would be great if folks could test this out and report back with confirmation of success or any issues. Please report back on the PR #2025 (or a new issue if any problems) as there are a few duplicate issues for this. To use a snapshot, `git pull && mvn install`, and use version `1.17.1-SNAPSHOT` in your project.

PR #1749, opened 2022-04-23
Fix #1731

### LINKED ISSUE #1731: Clean: Links with stripped href should  be removed
JSoup.clean() keeps links with stripped href like

```
<a href="invalid link">my link</a>

=> (JSoup.clean())

<a rel="nofollow">my link</a>
```

This makes no sense to me. The behaviour should be to strip the complete link to plaintext like

```
=>

my link
```

At least there should be a SafeList-Option to remove tags, if they do not have attribute xx after cleaning.

FKBugMaker:
Fix #1731


jhy:
Thanks. I'm not inclined to merge -- as noted on the bug, a elements can still be meaningful without href attributes.

Issue #1719, opened 2022-02-10
Inconsistent Handling Of Duplicate Attributes


alycecil:
Relates to https://github.com/jhy/jsoup/issues/1219 and https://github.com/jhy/jsoup/issues/1234

### Desired Behavior
Attribute duplicates other than the first should be ignored.

```
Scenario Outline: Jsoup handles duplicate consistently 
Given I parse the string "<img src=\"file.png\" name=\"test\" value=\"foo\" type=\"hidden\" value=\"bar\" />" as <From>
When I convert the document to <To>
Then I expect the html to contain "<img src=\"file.png\" name=\"test\" value=\"foo\" type=\"hidden\" />"
 Examples:
 |From|To|
 |HMTL|HTML|
 |HMTL|XML|
 |XML|XML|
 |Jsoup Direct|HTML|
 |Jsoup Direct|XML|
 ```
 
### Observed behavior
Dedupe only happens when `Parser.xmlParser()` consumes the input html.
 
### Example JUNIT 
```
import org.jsoup.Jsoup;
import org.jsoup.nodes.Document;
import org.jsoup.parser.Parser;
import org.junit.jupiter.api.Test;

import static org.assertj.core.api.Assertions.assertThat;

public class ConvertTest {
    private static final String HEAD = "<html>\n" +
            " <head></head>\n" +
            " <body>\n" +
            "  ";
    private static final String TRAIL = "\n" +
            " </body>\n" +
            "</html>";
    public static final String DESIRED_ONE_TAG = "<img src=\"file.png\" name=\"test\" value=\"foo\" type=\"hidden\" />";
    public static final String INPUT = "<img src=\"file.png\" name=\"test\" value=\"foo\" type=\"hidden\" value=\"bar\" />";
    public static final String INPUT_NO_SLASH = "<img src=\"file.png\" name=\"test\" value=\"foo\" type=\"hidden\" value=\"bar\">";

    @Test
    void parserXML() {
        String doubleTag = INPUT;
        Parser parser = Parser.xmlParser().setTrackErrors(10);
        Document doc = parser.parseInput(doubleTag, "");

        assertThat(doc.selectFirst("img").outerHtml()).isNotBlank().isEqualTo(DESIRED_ONE_TAG);
    }

    @Test
    void parserHTML() {
        String doubleTag = INPUT;
        Parser parser = Parser.htmlParser().setTrackErrors(10);
        Document doc = parser.parseInput(doubleTag, "");

[...]

PR #1750, opened 2022-04-23
FixBug of jsoup Inconsistent Handling Of Duplicate Attributes #1719

MVP-D77:
FixBug of jsoup Inconsistent Handling Of Duplicate Attributes #1719


jhy:
Thanks -- will close this per the review and note on #1719.

PR #1754, opened 2022-04-23
Solve testfailure when network is unable

### LINKED ISSUE #1733: ConnecTest#handlesUnknownEscapesAcrossBuffer fails if the network is not available
The test fails with an unknown host exception when the network is not available.
`
[ERROR] Errors: 
[ERROR]   ConnectTest.handlesUnknownEscapesAcrossBuffer:524 » UnknownHost gist.githubuse...
[INFO] 
[ERROR] Tests run: 1008, Failures: 0, Errors: 1, Skipped: 2
`

JeffXiesk:
Fix #1733 
As a software engineer, I want to use Jsoup library offline for some file parsing so that it is necessary for Jsoup to pass the test when offline.

### DIFF
--- a/src/main/java/org/jsoup/helper/Validate.java
+++ b/src/main/java/org/jsoup/helper/Validate.java
@@ -1,6 +1,8 @@
 package org.jsoup.helper;
 
 import javax.annotation.Nullable;
+import java.net.URL;
+import java.net.URLConnection;
 
 /**
  * Simple validation methods. Designed for jsoup internal use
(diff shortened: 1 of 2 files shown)


src/main/java/org/jsoup/helper/Validate.java
```diff
@@ -119,4 +121,22 @@ public static void wtf(String msg) {
     public static void fail(String msg) {
         throw new IllegalArgumentException(msg);
     }
+
+    /**
+     * Validate the connection
+     * @param url url needs to be connected
+     * @return if the connection is OK
+     */
+    public static boolean check_connection(String url) {
```
jhy, line 130: Why would this be a new public contract API? If we were to use it, would be better in a test helper class.

And please use normal Java style guide. This is not Python.

jhy:
Thanks. Feel it would be better just to move that specific test to the internal hosted tests. Not inclined to merge.

PR #1757, opened 2022-04-24
Find reason of null

### LINKED ISSUE #1756: the addition imprrove for the selectXpath function
in the closed issues "Question about using document.selectXpath() in xhtml #1746" Once XML with namespace is encountered, the corresponding XPath query will have no results. That is not a good thing for packet because it is not a easy thing to know is there XML with namespace in the http file. so my term and I plan to add a judge in it to throw a warning when the this case happen. I think this is important because no one want to get a null without the error message.

QAQGaeBolg:
fix #1756

### DIFF
--- a/src/main/java/org/jsoup/nodes/NodeUtils.java
+++ b/src/main/java/org/jsoup/nodes/NodeUtils.java
@@ -32,6 +32,31 @@ final class NodeUtils {
     }
 
     /**
+     * Judge whether the input document has a xmlns property,
+     * which will cause the selectXpath method return null.
+     * @param wDoc document to be judged.
+     * @param contextNode node to be judged.
+     * @param nodeList nodeList to be judged.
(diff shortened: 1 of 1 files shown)


jhy:
Thanks, but I don't see this as production ready. I don't know that it's a feature we need, and in this state with printlines it's not a meaningful API. Not inclined to merge.

jhy:
Thanks, but I don't see this as production ready. I don't know that it's a feature we need, and in this state with printlines it's not a meaningful API. Not inclined to merge.

susJohnny:
Sorry to bother you, but this is part of our course project. We don't mean to make any change to the code, just to leave the record and get scores. Could you please keep the commit and pull request record?

jhy:
Thanks for wasting our time on your point scoring, really appreciate it.

sol1775:
No me deja entrar

Issue #1724, opened 2022-02-18
XML Escape mode should escape > in attributes


ericapradoc:
The issue for '<' was solved in [#1420](https://github.com/jhy/jsoup/issues/1420) but '>' still not being escaped:

final Document doc = Jsoup.parse("<input class=\"text\" value=\" &gt; data  &lt; \"/>");
        doc.outputSettings().escapeMode(org.jsoup.nodes.Entities.EscapeMode.xhtml);
        System.out.println("Output: " + doc.select("input.text").get(0));

Current Output: `<input class="text" value=" > data  &lt; ">`
Expected Output: `<input class="text" value=" &gt; data  &lt; ">`

QAQGaeBolg:
I along with my group will be fixing this issue in this semester.

jhy:
@ericapradoc what makes you see this as an issue? `>` does not need to be escaped in XML attributes. See the spec here: https://www.w3.org/TR/xml/#syntax

Note that `>` _may_ (not _must_) be escaped.

There's a good summary here: https://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents/46637835#46637835

PR #1758, opened 2022-04-24
Fix to escape > in xhtml mode

LoveCany:
As mentioned in #1724. 

In my view, it is not necessary for xml syntax, so it should only escape in xhtml mode.


jhy:
Thanks, but > is OK in attributes in XML / XHTML. See the bug.

PR #1759, opened 2022-04-24
Add a new method for querying the n th matching elements / fix #1104

### LINKED ISSUE #1104: why ":nth-of-type(n)" doesn't support ?
HTML contents : 
jsoup

Here, I need to select "dl" tag with class=summary but it is present more than one time. I applied "dl:nth-of-type(1)[class=summary]" but it is selecting all dl tags with class value as summary.
Is there any other way to select a specific one out of similar elemnets ? 
Thanks

JeffXiesk:
fix #1104

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -423,6 +423,22 @@ public class Element extends Node {
     }
 
     /**
+     * Find the n th elements that match the query.
+     * @param n the n th element to be selected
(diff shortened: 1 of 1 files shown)


jhy:
Thanks, but I don't see this as a fit for jsoup. For one, that's not what nth-of-type means (see [ref](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type)), and it's already effectively implemented in the `:nth-of-type` selector. And this impl is just an odd alias for Elements#eq.

PR #1762, opened 2022-05-05
Adding a function for connection for more attempt / fix #1761

### LINKED ISSUE #1761: Add attempt
Hi!
You can add to the call parameters the number of attempts to receive an answer from the server.
**.attemp(3)**
This is necessary in order to make several attempts to receive an answer from the server if it is not equal to Code 200
It will be convenient not to create extra cycles.
```
			Document doc = Jsoup.connect(url)
                                        .attemp(3)
					.get();
```
Forgive my bad English.

JeffXiesk:
Fix #1761. Forgive for my poor coding ability.

### DIFF
--- a/src/main/java/org/jsoup/Jsoup.java
+++ b/src/main/java/org/jsoup/Jsoup.java
@@ -90,6 +90,39 @@ public class Jsoup {
     public static Connection connect(String url) {
         return HttpConnection.connect(url);
     }
+    
+    /**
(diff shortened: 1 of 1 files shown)


jhy:
No thanks - if we are going to add a retry facility, we should do it with something like a RetryHandler interface to include support for backoff, jitter, total timeout, etc.

Also, this commit is not production ready - it has no tests, is formated like Python and not Java, etc.

PR #1763, opened 2022-05-05
Fix/safelist deep copy constructor

### LINKED ISSUE #1764: Copy constructor of Safelist shares state
The copy constructor of Safelist `public Safelist(Safelist copy)` does not deep copy the data structures as advertised in the JavaDoc but shares the nested data structures, e.g. `Map<TagName, Set<AttributeKey>>`. This causes unexpected mutation.

An example would be:
```java
Safelist safelist1 = Safelist.none().addAttributes("foo", "bar");
Safelist safelist2 = new Safelist(safelist1);
safelist1.addAttributes("foo", "baz");

final String html = "<foo bar=\"val\" baz=\"val\" />";
System.out.println(Jsoup.clean(html, safelist1));
System.out.println(Jsoup.clean(html, safelist2));
```
The second output should not contain the `baz` attribute but does.

cschwier:
Fixes a bug where the copy constructor of Safelist would share state with its source Safelist which can unexpectedly mutate both safelists if the modifying operation is invoked on either.
In particular, this affected the allowed attributes, enforced attribute values and protocols.

fixes #1764

### DIFF
--- a/src/main/java/org/jsoup/safety/Safelist.java
+++ b/src/main/java/org/jsoup/safety/Safelist.java
@@ -203,9 +203,19 @@ public class Safelist {
     public Safelist(Safelist copy) {
         this();
         tagNames.addAll(copy.tagNames);
-        attributes.putAll(copy.attributes);
-        enforcedAttributes.putAll(copy.enforcedAttributes);
-        protocols.putAll(copy.protocols);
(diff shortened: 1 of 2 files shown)


jhy:
Looks great, thanks Christian. Merged.

PR #1779, opened 2022-05-25
Update Element.java and add NewElementTest.java

### LINKED ISSUE #1117: Showing content of downlevel-revealed conditionals for IE
I'm parsing with Jsoup the follwing html and the output html and nodes are different in the result of `document.html()`
```        
String html = "<div>"
            + "<![if true]>1<![endif]>" // downlevel-revealed
            + "<!--[if true]>2<![endif]-->" // downlevel-hidden
            + "</div>";
Document document = Jsoup.parse(html);
```
**Output**

```
<html>
 <head></head>
 <body>
  <div>
   <!--[if true]-->1
   <!--[endif]-->
   <!--[if true]>2<![endif]-->
  </div>
 </body>
</html>
```
**Expected**

```
<html>
 <head></head>
 <body>
  <div>
   <![if true]>1<![endif]>
   <!--[if true]>2<![endif]-->
  </div>
 </body>
</html>
```
And the div `document.select("div").first().childNodes()` are like this:

class org.jsoup.nodes.Comment
class org.jsoup.nodes.TextNode
class org.jsoup.nodes.Comment
class org.jsoup.nodes.Comment

I would be happy to contribute to an eventual patch if you give me some hints.

FKBugMaker:
Fix #1117

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -1654,19 +1654,31 @@ public class Element extends Node {
         }
     }
 
-    /**
-     * Retrieves the element's inner HTML. E.g. on a {@code <div>} with one empty {@code <p>}, would return
-     * {@code <p></p>}. (Whereas {@link #outerHtml()} would return {@code <div><p></p></div>}.)
-     *
-     * @return String of HTML.
-     * @see #outerHtml()
-     */
-    public String html() {
-        StringBuilder accum = StringUtil.borrowBuilder();
-        html(accum);
-        String html = StringUtil.releaseBuilder(accum);
-        return NodeUtils.outputSettings(this).prettyPrint() ? html.trim() : html;
+  /**
+   * Retrieves the element's inner HTML.
+   * E.g. on a {@code <div>} with one empty {@code <p>}, would return
+   * {@code <p></p>}. (Whereas {@link #outerHtml()} would return {@code <div><p></p></div>}.)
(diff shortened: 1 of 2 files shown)


jhy:
Thanks, but no thanks. The serialization methods are not the time to be introspecting the DOM. The parsed DOM should reflect jsoup's view of the data. Apart from scattering parse logic all throughout the codebase, this implementation is going to be a significant performance impact.

As I mentioned in the bug, I don't know that the feature request of #1117 can be implemented in a HTML5 standards compliant way. If we were to add it (and again, I'm not convinced that there's sufficient need/utility), I think it would have to be as an optional extra non-standard tag support feature. And if we were doing that, it would be best to do it in a generic way so that users could parse for e.g. PHP or other non-standard tags.

FKBugMaker:
> Thanks, but no thanks. The serialization methods are not the time to be introspecting the DOM. The parsed DOM should reflect jsoup's view of the data. Apart from scattering parse logic all throughout the codebase, this implementation is going to be a significant performance impact.
> 
> As I mentioned in the bug, I don't know that the feature request of #1117 can be implemented in a HTML5 standards compliant way. If we were to add it (and again, I'm not convinced that there's sufficient need/utility), I think it would have to be as an optional extra non-standard tag support feature. And if we were doing that, it would be best to do it in a generic way so that users could parse for e.g. PHP or other non-standard tags.

OK. I know what you mean. Thanks.

Issue #1278, opened 2019-11-13
Expose a public method to escape characters within an attribute


ghost:
Hello

I need to escape the characters within an attribute in a program that converts an HTML file into a single Java class with a single public static method that converts the Java source code into HTML:
https://github.com/gouessej/HtmlFlow/blob/patch-1/src/main/java/htmlflow/flowifier/AbstractHtmlToJavaHtmlFlowNodeVisitor.java#L319

[The package protected static method named "escape"](https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/nodes/Entities.java#L173) does the correct job by passing inAttribute at true but I can't call it in the source code above because it's not public. Please can you provide a public method that allows to pass inAttribute at true? It would be cleaner than my dirty kludge that doesn't take into account some other cases.

I suggest the addition of the following method:
   ```
/**
     * HTML escape an input string within an attribute.
     *
     * @param string the un-escaped string within an attribute to escape
     * @param out the output settings to use
     * @return the escaped string
     */
    public static String escapeInAttribute(String string, Document.OutputSettings out) {
        if (string == null)
            return "";
        StringBuilder accum = StringUtil.borrowBuilder();
        try {
            escape(accum, string, out, true, false, false);
        } catch (IOException e) {
            throw new SerializationException(e); // doesn't happen
        }
        return StringUtil.releaseBuilder(accum);
    }
```

Best regards and keep up the good work.

mulinJiao:
In jsoup, using `Parser.xmlParser()` to generate the DOM is not the same as using `Parser.htmlParser()`.
E.g.
tagName in `Parser.xmlParser()` - case sensitive 
tagName in `Parser.htmlParser()` - lower case.

And when we use the selector - `*|MyXmlTag` - the string is converted to lowercase.
Therefore, in the case `Parser.xmlParser()`, the selector will find nothing.

jhy:
Thanks for the suggestion - I have updated `Entities.escape(string)` so that it's output is suitable for both text data and attributes. It's more conservative or maximal. The output from `Element.html()` etc is still more conservative and uses the context of if in an attribute or in text to decide what to escape.

PR #1782, opened 2022-05-28
fix issue 1278

rgseyvie:
Add public method to escape characters within an attribute.
Issue link: https://github.com/jhy/jsoup/issues/1278


jhy:
I don't get why you're re-implementing the whole escape method. That duplicates a ton of code which is wasteful and brittle. Why not just create a new method and call `escape` with inAttribute to be `true`?

Issue #1767, opened 2022-05-12
Allow HTML5 details tag in Safelist.relaxed


dschulten:
Would it be possible to include the HTML5 Tags `details` and `summary` in `Safelist.relaxed`?

https://developer.mozilla.org/de/docs/Web/HTML/Element/details

Or are there reasons why those two tags cannot be considered safe?

jeffthomasweb:
Apologies if this is already widely known, but if anyone would like to add their own safe HTML tags for current Jsoup versions, Safelist.addTags("stringTag1","stringTag2") is available. I've included an example below. https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/safety/Safelist.java#L222
```
String cleanRelaxedIncludeHtmlDetailsSummary() throws IOException {
        Safelist listOfSafeHtml = Safelist.relaxed().addTags("details","summary");
        String htmlWithDetailsSummary = "<details><summary>Details</summary>Something small enough to escape casual notice.</details>";
        String afterCleanedRelaxedHtml = Jsoup.clean(htmlWithDetailsSummary, listOfSafeHtml);

        return afterCleanedRelaxedHtml;
    }
```

jhy:
Tracking adding HTML5 tags to the canned lists in the #2284 Cleaner and Safelist revamp.

PR #1784, opened 2022-05-28
Implement Html5 safelist

LoveCany:
As mentioned in #1767, I think it necessary to add safelist for HTML5 tags.

Now only some tags useful for myself and those mentioned in that issue are implemented in this PR. Please point out more and I will add them.


rkovarik:
Could you add the "<s>" tag? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike

Issue #750, opened 2016-08-20
Add new method org.jsoup.nodes.Document#xml()


ghost:
Hi,

I know there's a method `Element#html()` to generate html. However xml has a header like:

``` xml
<?xml version="1.0" encoding="utf-8"?>
```

and `Element#html()` won't generate that header, which is true, of course.

So new method `Document#xml()` would come in handy. I would suggest to put it in `Document`, because to generate an element (without any header like above), we already have `Element#html()`.

It would be great if you consider supporting this.

Thank you,

jhy:
Calling [doc.charset(UTF8)](https://jsoup.org/apidocs/org/jsoup/nodes/Document.html#charset(java.nio.charset.Charset)) is just about equivalent to this, so not currently planning to add this.

PR #1785, opened 2022-05-29
Issue #750 fixed by adding a Document#xml() and two testcases

playmakerv6:


jhy:
Not really sure on the point of this PR, it's very half baked. What happens when the DOM already has a an XML declaration? What if the encoding is not UTF-8?

There are already XML syntax output methods. Can you clarify what problem you are trying to actually address?

Issue #1378, opened 2020-05-15
Add Node#noteType() and constants?


furstenheim:
Hi,
Very nice library. It seems that there is no way to get NodeType in the API.
https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType
Thanks.

jhy:
We could potentially add it. There is `nodeName` which returns a unique string. Generally I've just used the class instanceOf.

windsliver:
I'm interested in working on this issue.

jhy:
Closing, not planned. I would just use instanceof.

PR #1786, opened 2022-05-29
Fix issue #1378 by add  Node#noteType()

windsliver:
Use Node#noteType() to get NodeType easily.


jhy:
This seems half baked. There's no actual tests -- do you think people are going to read and check them on every test run?

And if there is a missed node it returns 1. Seems like that should blow up instead to identify any gaps early.

And I think the impl, if we actually need it, would be better with the constants in the relevant node class.

I'm not clear on the utility of this method. Do you actually need it? Why is it better than using `instanceof`? 

Or are you just generating PRs for point scoring in your class?

PR #1793, opened 2022-06-15
Correction information about InputStream closure

jartysiewicz:


jhy:
Good stuff, thanks! Merged. 

I couldn't see that IntelliJ actually did anything with the annotation. Changing it to `@WillNotClose` did not flag a warning. I'd be interested to know if other editors do have an implemented test. Regardless, it's informative to include.

jhy:
Unfortunately as part of #2048 I needed to remove the @ WillClose annotation, and jspecify doesn't offer a replacement.

Open to ideas here!

PR #1832, opened 2022-08-31
Proposed new settings to leave HTML entities untouched and to clean the contents of attribute values

kworam-rally:
Some proposed modifications to jsoup:

1. Introduce Entities.EscapeMode.none which causes Entities.escape() to leave HTML entities untouched.
2. Introduce Cleaner.CleanerSettings with cleanAttributeValues() and baseUri(). If cleanAttributeValues() is true, the Cleaner cleans HTML embedded in attribute values using baseUri().


jhy:
I will never implement an None entities escape mode in jsoup. It is a loaded foot-gun, and indicative that the library is being used incorrectly. If you want plain text output, use one of the .text() methods. If one of those doesn't fit the use case, I'd be happy to hear more about the use case and we can explore ways to improve those.

PR #1835, opened 2022-09-07
Change simple cookie header parse to keep most recently set

### LINKED ISSUE #1831: Simple cookie parser keeps first cookie with same name, not most recently set
image

lvpiao:


jhy:
Thanks, and sorry for the late review. I have implemented the fix for #1831 in b6bd4b455c54fa44c68df961b9d5c21a2bb0ae73. This works correctly if there are multiple cookies with the same name but different paths - just going with the most recently set may not be the correct choice in that case.

anonyein:
I recommend that this commit broken "set-cookie" list and makes cookies lacks of lost of messages!
https://github.com/jhy/jsoup/commit/b6bd4b455c54fa44c68df961b9d5c21a2bb0ae73

PR #1844, opened 2022-09-22
Proposed fix : Disallow replacing nodes with themselves

### LINKED ISSUE #1838: Jsoup 1.14.3 & 1.15.3 -> org.jsoup.nodes.Element : Self closing tags must not have children but Jsoup allows them to have children.
Hi @jhy ,
Thanks for reviewing this issue.
Sorry, I changed the title because it was confusing. 
Yes, HTML5 does not require a trailing / to make an empty tag. 
I was trying to explain the problem we faced with Jsoup API.

**Our goal is not to parse HTML to get a parsed HTML as output  as it is done here https://try.jsoup.org/ but we are using Jsoup to turn HTML elements into Java Objects (  [Node](https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/nodes/Node.java), [Element](https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/nodes/Element.java) and [TextNode](https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/nodes/TextNode.java) from Jsoup ) then our app turns these Java Objects  into JavaScript variables.**

**While working, we noticed that if there are 2 self closing tags following each other then the 2nd one is considered as the child of the 1st one. To be more precise, we also noticed that when the 1st one has slash at the end then everything is correct but when the 1st one doesn't have slash, we got this issue. After so many trials, we concluded that every tag following a self closing tag without slash is considered as its child.** 

 **It is impossible to see that here https://try.jsoup.org/ . The only way to see that is to deeply analyze the code to get a better idea on how things work under the hood.**

So, how  to reproduce this issue ?

Just run the code below, we invite you to use the 3 strings (selfClosingTagsWorking, selfClosingTagNotWorking and brDiv) to see by yourself. Also run with many examples as you want to better understand.

```java
import org.jsoup.Jsoup;
import org.jsoup.nodes.Element;
import org.jsoup.parser.Parser;

public class SelfClosingTagIssue {

	String selfClosingTagWorking = "<input type=\"text\"/>\r\n" + "<img src=\"#URL\" alt=\"image\">";

	String selfClosingTagNotWorking = "<input type=\"text\">\r\n" + "<img src=\"#URL\" alt=\"image\">";

	String brDiv = "<br> <div><div/>";

	Element htmlDoc = Jsoup.parse(brDiv, "", Parser.xmlParser());

	for (Element child : htmlDoc.children()) {

	    boolean hasChild = child.childrenSize() > 0;

	    System.err.println(" **** Analyze :" + child.tag().getName() + " -> isEmpty : " + child.tag().isEmpty()
		    + " -> isSelfClosing : " + child.tag().isSelfClosing() + " -> hasChild : " + hasChild + " **** ");

	    if (hasChild) {
		System.err.println("Children of " + child.tagName() + ": ");

		for (Element e : child.children()) {

		    System.err.println(e.tagName());

		}
	    }

	}

    }

}

```

Result with selfClosingTagWorking

```
 **** Analyze :input -> isEmpty : true -> isSelfClosing : true -> hasChild : false **** 
 **** Analyze :img -> isEmpty : true -> isSelfClosing : true -> hasChild : false **** 

```

Result from selfClosingTagNotWorking

```
 **** Analyze :input -> isEmpty : true -> isSelfClosing : true -> hasChild : true **** 
Children of input: 
img

```

Result from brDiv

```
 **** Analyze :br -> isEmpty : true -> isSelfClosing : true -> hasChild : true **** 
Children of br: 
div

```

_Originally posted by @FanJups in https://github.com/jhy/jsoup/issues/1741#issuecomment-1246952249_

### LINKED ISSUE #1843: Replacing an element with itself causes problems.
Obviously there's no practical use in replacing an element with itself, but robustness is in the edge cases.

Take the document: 
```
body
| h1
| div
```

If we were to replace the h1 with itself via `h1.replaceWith(h1)`, it's next sibling, in this case the div, would be deleted and the h1 would remain - but without a reference to it's parent.

In this state, the node traverser fails when rendering the tostring().

Minimal reproducable :

```Kotlin
        val doc = Document("")

        val targetElement = Element("h1")
        doc.body().appendChild(targetElement)

        // During the replace, the target is removed from the children.
        // 'Ensure child list' within the replace operation is
        // different if there is no children, so we must add another child.
        //
        // Also, there must be children *after* the target element, else
        // there will be an index out of bounds during the replace operation.
        doc.body().appendChild(Element("div"))

        targetElement.replaceWith(targetElement)

        // The document can no longer be rendered to string.
        assertDoesNotThrow {
            println(doc)
        }
```

This can be solved with a simple equality check.

Self assigned, pr soon

jdngray77:
fixes #1843 : `Node.replaceWith` deletes siblings and creates an invalid state if a node is replaced with itself.

Proposed changes :
 * Equality validation, throw a `ValidationException` just as any other invalid parameter would.
 * Unit test for edge case

### DIFF
--- a/src/main/java/org/jsoup/helper/Validate.java
+++ b/src/main/java/org/jsoup/helper/Validate.java
@@ -10,6 +10,18 @@ public final class Validate {
     private Validate() {}
 
     /**
+     * Validates that the two objects are not the same.
+     * @param a First object
+     * @param b Second object
+     * @param message Error message to use if the objects are the same
+     * @throws IllegalArgumentException if the objects are the same
+     */
+    public static void notStrictlyEqual(Object a, Object b, String message) {
+        if (a == b)
+            throw new ValidationException(message);
+    }
+
+    /**
      * Validates that the object is not null
      * @param obj object to test
      * @throws ValidationException if the object is null
--- a/src/main/java/org/jsoup/nodes/Node.java
+++ b/src/main/java/org/jsoup/nodes/Node.java
@@ -457,6 +457,7 @@ public abstract class Node implements Cloneable {
     public void replaceWith(Node in) {
         Validate.notNull(in);
         Validate.notNull(parentNode);
+        Validate.notStrictlyEqual(this, in, "Cannot replace a node with itself");
         parentNode.replaceChild(this, in);
     }
 
--- a/src/test/java/org/jsoup/nodes/NodeTest.java
+++ b/src/test/java/org/jsoup/nodes/NodeTest.java
@@ -2,6 +2,7 @@ package org.jsoup.nodes;
 
 import org.jsoup.Jsoup;
 import org.jsoup.TextUtil;
+import org.jsoup.helper.ValidationException;
 import org.jsoup.parser.Tag;
 import org.jsoup.select.NodeVisitor;
 import org.junit.jupiter.api.Test;
@@ -401,5 +402,26 @@ public class NodeTest {
 
         assertNull(firstEl.firstElementChild());
         assertNull(firstEl.lastElementChild());
+    }
+
+    @Test
+    void replaceWithSelf() {
+        // Issue 1843
+
+        Document doc = new Document("");
+        Element h1 = doc.body().appendElement("h1");
+        Element div = doc.body().appendElement("div");
+
+        assertDoesNotThrow(doc::toString);
+
+        assertThrowsExactly(ValidationException.class, () -> h1.replaceWith(h1));
+
+        assertTrue(doc.body().children().contains(h1));
+        assertTrue(doc.body().children().contains(div));
+        assertNotNull(h1.parent());
+        assertNotNull(div.parent());
+        assertDoesNotThrow(doc::toString);
+
+
     }
 }


FanJups:
Thanks for this correction 💯

jdngray77:
My apologies, I typed the wrong issue number - this is unrelated to your problem of self closing tags.

FanJups:
all right lol

src/main/java/org/jsoup/helper/Validate.java
```diff
@@ -9,6 +9,18 @@ public final class Validate {
...
     private Validate() {}
 
+    /**
+     * Validates that the two objects are not the same.
+     * @param a First object
+     * @param b Second object
+     * @param message Error message to use if the objects are the same
+     * @throws IllegalArgumentException if the objects are the same
+     */
+    public static void notStrictlyEqual(Object a, Object b, String message) {
```
jhy, line 19: Personally, I would call this method something like "NotSameObject". To make it clearer that it's an identity check.

jhy:
Thanks for the PR and the bug report.

I think that it would be better to make this a no-op vs a validation error. It follows closer to the developer's intent (whatever that was) and is not surprising. 

I.e., if `(this == in) return`

jhy (changes requested):
As noted inline -- would be better if it was a no-op vs a Validation Error.

jhy:
I implemented this with 9bb07d2ab43c0377806b4bb7b88c64366a39a469

Issue #1811, opened 2022-07-21
"ns|* all elements in namespace ns" selector is not supported


rezss:
Hey,

the [ns|* selector](https://www.w3.org/TR/selectors-3/#univnmsp) unfortunately doesn't find any elements.

```kotlin
@Test
fun `Can find all ac namespace elements`() {
    val html = """<html><body><p>Hello, World!</p><ac:p>p in ac namespace</ac:p><ac:img>img in ac namespace</ac:img><ac:page>page in ac namespace</ac:page></body></html>"""
    val document = Jsoup.parse(html)

    val namespaceElements = document.select("ac|*")

    assertThat(namespaceElements.size).isEqualTo(3)
    // expected: 3
    // but was: 0
}
```

Tried with **org.jsoup:jsoup:1.15.2**

jhy:
Thanks - will be good to add support for this selector.

mo20053444:
@jhy Hi! Can I try this? Thanks!

mo20053444:
@jhy Hi! I created a PR for this implementation [(link)](https://github.com/jhy/jsoup/pull/1847#issue-1393422851)
Would you mind taking a look?
Many thanks!

jhy:
Thanks, I've implemented this now.

PR #1847, opened 2022-10-01
Add selecting elements by namespace (#1811)

mo20053444:
This commit enables users to select all elements within a particular namespace by inputting queries like "ns|*".


src/main/java/org/jsoup/select/QueryParser.java
```diff
@@ -162,6 +162,8 @@ private void findElements() {
             byId();
         else if (tq.matchChomp("."))
             byClass();
+        else if (tq.toString().endsWith("|*"))
```
jhy, line 165: This is not the right way to do this. The TokenQueue must be consumed token by token. With this implementation, queries like `ns|* div` (or anything after the namespace wildcard selector) will fail. You will need an appropriate matcher in TokenQueue. I think you could update the current tag matcher.

src/main/java/org/jsoup/select/QueryParser.java
```diff
@@ -262,6 +264,12 @@ private void byTag() {
         }
     }
 
+    private void byNamespace() {
```
jhy, line 267: I think I would just extend the byTag method to support this.

src/test/java/org/jsoup/nodes/ElementTest.java
```diff
@@ -1319,6 +1319,16 @@ public void testNamespacedElements() {
         assertEquals("html > body > fb|comments", els.get(0).cssSelector());
     }
 
+    @Test
+    public void testSelectByNamespace() {
```
jhy, line 1323: This should be in SelectorTest, and needs to be more detailed -- e.g. to catch the example I pointed out, and validate other variations work.

jhy (changes requested):
Review notes inline.

mo20053444:
@jhy Hi Jonathan, hope you had a great holiday season and thank you for your feedback! I will take a look at your comments and make changes accordingly. Thanks!

Issue #1837, opened 2022-09-13
Bug: DOM elements not being placed in (X)HTML namespace.


garretwilson:
The [description page](https://jsoup.org/) for jsoup says:

> jsoup implements the [WHATWG HTML5](https://whatwg.org/html) specification, and parses HTML to the same DOM as modern browsers do.

The [WHATWG HTML5 Specification § 2.1.3 XML compatibility](https://html.spec.whatwg.org/multipage/infrastructure.html#xml) says:

> To ease migration from HTML to XML, user agents conforming to this specification will place elements in HTML in the [http://www.w3.org/1999/xhtml](https://infra.spec.whatwg.org/#html-namespace) namespace, at least for the purposes of the DOM and CSS.

In other words, jsoup should be placing HTML elements in the `http://www.w3.org/1999/xhtml` namespace, even in the absence of an `xmlns` declaration. But it's not.

I'm using `org.jsoup:jsoup:1.15.3` with Java 17. I have a test HTML document that looks something like this:

```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <title>Foobar</title
</head>
<body>
  …
</body>
```

In a JUnit 5/Hamcrest unit test I read that document:

```java
URL testResourceUrl = getClass().getResource("foobar.html");
try (InputStream inputStream = testResourceUrl.openStream()) {
  org.jsoup.nodes.Document jsoupDocument = Jsoup.parse(inputStream, null, testResourceUrl.toString());
  org.w3c.dom.Document domDocument = new W3CDom().fromJsoup(jsoupDocument);
  assertThat(domDocument.getDocumentElement().getNamespaceURI(), is("http://www.w3.org/1999/xhtml"));
}
```

Unfortunately this test fails:

```
java.lang.AssertionError: 
Expected: is "http://www.w3.org/1999/xhtml"
     but: was null
…
```

Thus jsoup has a bug: it is not following the HTML5 specification and placing HTML elements in the `http://www.w3.org/1999/xhtml` namespace, but instead assigning them the `null` namespace.

To reiterate, yes, I realize that the HTML document in question has no default namespace specified. But as an HTML document it isn't required or even expected to. The HTML5 specification says that the HTML elements should be placed in the `http://www.w3.org/1999/xhtml` regardless.

garretwilson:
[...]

PR #1848, opened 2022-10-01
Put W3C DOM elements in HTML namespace by default for jhy/jsoup#1837.

garretwilson:
Fixes jhy/jsoup#1837.


src/main/java/org/jsoup/helper/W3CDom.java
```diff
@@ -366,9 +406,9 @@ public void head(org.jsoup.nodes.Node source, int depth) {
                 tagname to something safe, because that isn't going to be meaningful downstream. This seems(?) to be
                 how browsers handle the situation, also. https://github.com/jhy/jsoup/issues/1093 */
                 try {
-                    Element el = namespace == null && tagName.contains(":") ?
-                        doc.createElementNS("", tagName) : // doesn't have a real namespace defined
-                        doc.createElementNS(namespace, tagName);
+                    // use an empty namespace if none is present but the tag name has a prefix
```
garretwilson, line 409: I didn't change any functionality here. The modification makes it clear that the only thing happening is that you're "imputing" a namespace of the empty string under certain conditions — you're not calling a separate method. The existing code duplicated the call to `doc.createElementNS()`, which primarily was confusing and obscured the purpose of the logic.

garretwilson:
I don't know why some of the integration tests are failing on macOS, but from the output it doesn't appear to be related to any changes here. My changes shouldn't have anything to do with integration-level issues, and the unit tests are all working.

src/main/java/org/jsoup/helper/W3CDom.java
```diff
@@ -348,6 +387,7 @@ protected static class W3CBuilder implements NodeVisitor {
         public W3CBuilder(Document doc) {
             this.doc = doc;
             namespacesStack.push(new HashMap<>());
+            namespacesStack.peek().put("", "http://www.w3.org/1999/xhtml"); // TODO document
```
jhy, line 390: I feel that we should only be setting this default namespace if the input document is HTML. If it's XML, then the HTML -> XML compat note doesn't apply, and we shouldn't be applying it.

The impl could be something like
```
if (inDoc.parser().getTreeBuilder() instanceof HtmlTreeBuilder)
```
Would need to move this to the appropriate `convert()` method. Or, add that flag as user data to the W3C Document.

I'm not sure if checking the treebuilder is the best way to see if it's HTML vs XML. The output syntax is how we normally check that, but that doesn't tell us particularly if the input was HTML, and people may be setting the syntax to XML before calling these methods. So am leaning towards checking treebuilder.

src/main/java/org/jsoup/helper/W3CDom.java
```diff
@@ -96,6 +97,35 @@ public static Document convert(org.jsoup.nodes.Document in) {
...
+     * @param html The serialized HTML.
+     * @return A string without the default <code>xmlns="http://www.w3.org/1999/xhtml"</code> HTML
+     *         namespace declaration.
+     * @see <a href="https://github.com/jhy/jsoup/issues/1837">Issue #1837: Bug: DOM elements not
+     *      being placed in (X)HTML namespace.</a>
+     */
+    static String removeDefaultHtmlNamespaceDeclaration(String html) {
+        Matcher matcher = HTML_DEFAULT_NAMESPACE_PATTERN.matcher(html);
+        if (matcher.find()) {
+          html = html.substring(0, matcher.start(1)) + html.substring(matcher.end(1));
```
jhy, line 124: I think it is better to leave the introduced namespace in the output. It makes the output XML clearer as to what's happened, and may help other downstream parses. So the regex components etc can be removed.

> We can leave it like that or play tricks to get rid of it, but it depends on the purpose and usage of W3CDom.asString(). Is it just used for testing? Or is it part of the jsoup API for turning DOM into a string?

It's part of the API.

> jsoup is about parsing, not pretty-printing

I consider jsoup to be also about pretty-printing. And we do introduce other tokens throughout jsoup as a convenience (e.g. xml declarations, meta charsets, the HTML document structure, etc).

jhy (changes requested):
Thanks for the PR. Please see a couple notes inline and let me know if you have any thoughts.  

LMK if you want to proceed with the changes, otherwise I am happy to apply them also.

jhy:
> I don't know why some of the integration tests are failing on macOS, but from the output it doesn't appear to be related to any changes here. My changes shouldn't have anything to do with integration-level issues, and the unit tests are all working.

Yep don't worry about those - likely to have just been GitHub not having enough Mac resources on hand when the tests ran to complete some of the timing tests within the limit.

src/main/java/org/jsoup/helper/W3CDom.java
garretwilson: > I think it is better to leave the introduced namespace in the output.

For my purposes I don't care one way or another, and it's certainly easier just to leave it in. At the time, due to lack of response to the ticket, I had to make a choice and do what I thought would raise the chance of the PR being requested. (You can see all the time and painful [research on Stack Overflow](https://stackoverflow.com/q/73919306) to remove the declaration.) I'll remove the code that removed the declaration in the output.

src/main/java/org/jsoup/helper/W3CDom.java
garretwilson: > I feel that we should only be setting this default namespace if the input document is HTML.

That makes sense. I don't know enough about your library's API to know when it thinks it's parsing XML. I had assumed it only parsed things considered to be non-XML HTML. (Otherwise I would think you could just use an off-the-shelf XML parser because the input would not be "soup".)

> So am leaning towards checking treebuilder.

You know best here. All I care about is when I parse an HTML document, the HTML namespace gets imputed as per the space. If the tree builder is always set to `HtmlTreeBuilder` in those cases, that works. Whatever you want to do.

> Would need to move this to the appropriate `convert()` method. Or, add that flag as user data to the W3C Document.

That sounds like a lot of shuffling. I note that the `W3CBuilder` does seem to get passed the `org.jsoup.nodes.Element` being converted via the W3C document user data context property. In normal cases that element would know its owner `org.jsoup.nodes.Document` would it not? (If not, can we assume this is HTML input? I would imagine it to be a rare case anyway for the jsoup element not to have an owner document.)

So what about this inside the `W3CBuilder` constructor (where I made the change already)?

```java
final org.jsoup.nodes.Document inDoc = contextElement.ownerDocument();
if (inDoc == null || inDoc.parser().getTreeBuilder() instanceof HtmlTreeBuilder) {
  namespacesStack.peek().put("", "http://www.w3.org/1999/xhtml"); // TODO document
}
```

(I see that I had forgotten to document the actual namespace imputation line. I'll add that once I get an OK from you on how to proceed.)

garretwilson: @jhy could you respond here and let me know which way to go, so that I can finish this PR and get it accepted?

jhy: > That sounds like a lot of shuffling. I note that the `W3CBuilder` does seem to get passed the `org.jsoup.nodes.Element` being converted via the W3C document user data context property. In normal cases that element would know its owner `org.jsoup.nodes.Document` would it not? (If not, can we assume this is HTML input? I would imagine it to be a rare case anyway for the jsoup element not to have an owner document.)

OK, using the contextElement is a good idea. And yes let's go with checking if the builder was HtmlTreeBuilder. That will work for all cases in the core library (vs anyone's own extensions, which is kind of up them).

I would not assume that an Element without an ownerDocument is HTML though, so we should not add the namespace if it is null. There will be no ownerDoc for Elements constructed via `new` and not attached to another doc, which is a supported use case. And in that case we don't know either way.

I plan on making this namespace inference a defaulted-on option in the W3C DOM. This will allow existing uses who have working code without a namespace (for e.g. simple xpath queries) a migration path, or for new uses that prefer it otherwise.

garretwilson:
> OK, using the contextElement is a good idea. And yes let's go with checking if the builder was HtmlTreeBuilder. That will work for all cases in the core library (vs anyone's own extensions, which is kind of up them).

@jhy I've made the changes you've requested. Sorry for the delay. I merged in your latest changes from `master` so this pull request should now be completely up to date.

Can you can review this and merge it soon? I really don't want for it to get stale, and each time I come back to it after a couple of months it's harder to re-orient myself to where I left it. 😅 

Let me know if you need further changes.

jhy:
Thanks, merged! Appreciate your perseverance in getting it completed.

I added a test that the converter is in namespace aware mode before applying it, so that it can be optionally disabled.

garretwilson:
Woohoo!! Thanks; this is exciting.

PR #1871, opened 2022-12-23
Allow wildcards in whitelist attributes

promiselaoliu:
Support validation on data attributes.


jhy (changes requested):
I thought this was going to be about wildcards, is framed around a new method for data- / custom "data" attributes. attributes. On review of the issue I still lean towards a wildcard match. Like 'data-*' or 'ng-*' etc. I think that will be more flexible and simpler to use than another method. It can be implemented in the isSafeAttribute method, doesn't need another data structure to retain them,

promiselaoliu:
> I thought this was going to be about wildcards, but it is framed around a new method for data- / custom "data" attributes. attributes. On review of the issue I still lean towards a wildcard match. Like 'data-_' or 'ng-_' etc. I think that will be more flexible and simpler to use than another method. It can be implemented in the isSafeAttribute method, doesn't need another data structure to retain them,

Fixed.

jhy:
Sorry for the late reply. I'm not moving ahead with this, as I think the implementation can be simpler, as I suggested in my earlier comment.

PR #1875, opened 2022-12-31
Use StandardCharsets.

Isira-Seneviratne:
Use the `StandardCharsets` class, as it is backported for Android API levels below 19 with [core library desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring) starting with [`desugar_jdk_libs`](https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md) 2.0.0.


jhy:
Thanks, but with the signature check build failing, I won't merge it. I am also not sure there's any actual benefit, given that these are statically initialized and cached.

PR #1880, opened 2023-01-09
Convert Accumulator to a lambda.

Isira-Seneviratne:
Convert the private `Accumulator` class in `Collector` to a lambda expression.


jhy:
Love commits with negative total lines of code :) Thanks

PR #1884, opened 2023-01-13
Use ThreadLocal.withInitial().

Isira-Seneviratne:
Use [`ThreadLocal.withInitial()`](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html#withInitial-java.util.function.Supplier-) in `StringUtil`, as it is [supported](https://issuetracker.google.com/issues/160484830) via desugaring.


jhy:
Thanks @Isira-Seneviratne. At the moment my plan is not to bring this in until we have a release with the `Consumer` annotation out. I want to see if anything shakes out from that - and particularly as it's only used in an optional code path. Assuming everything is fine we can move to these in the next major release 1.16.1.

PR #1895, opened 2023-02-09
fix little missunderstanding

JamesHertz:
The documentation of [Document.postDataCharset(String charset)](https://jsoup.org/apidocs/org/jsoup/Connection.html#postDataCharset(java.lang.String)) says: "Sets the default post data character set for x-www-form-urlencoded post data." 

At first I thought, it was not about the POST request but about the x-www-form-urlencoded data. Which means, that if I used this method sending a charset B as argument, even if I do a GET request the x-www-form-urlencoded data in it would've been converted into chartset B.

What is actually happening is that it's all about POST request and somehow it feels a bit weird. What is happening is that when it's a POST request Jsoup uses the charSet specified by postDataCharSet to transform the  x-www-form-urlencoded  data but for everything else it just uses a default charSet. The most funny thing is that the variable that postDataCharSet affects is initialized with this default charset.

I think that it would be better if Jsoup just used this value (specified using postDataCharSet) to transform x-www-form-urlencoded data, no matter the method used (as long as the method allows such kind of data).

I just changed two lines of code to make that happen.

I hope you reconsider this, it would make some people life easiser.

### DIFF
--- a/src/main/java/org/jsoup/helper/HttpConnection.java
+++ b/src/main/java/org/jsoup/helper/HttpConnection.java
@@ -1240,9 +1240,9 @@ public class HttpConnection implements Connection {
                 else
                     first = false;
                 url
-                    .append(URLEncoder.encode(keyVal.key(), DataUtil.defaultCharsetName))
+                    .append(URLEncoder.encode(keyVal.key(), req.postDataCharset()))
                     .append('=')
-                    .append(URLEncoder.encode(keyVal.value(), DataUtil.defaultCharsetName));
+                    .append(URLEncoder.encode(keyVal.value(), req.postDataCharset()));
             }
             req.url(new URL(StringUtil.releaseBuilder(url)));
             req.data().clear(); // moved into url as get params


charlie-smma (approved):

jhy:
Hi -- x-www-form-urlencoded data is the way we encode form data in the request body of a post. So the documentation and action are correct - the setting impacts the encoding of the request body. form-urlencoded in UTF8 really just means fields are split with `&` and `=` but can contain non ascii without being % encoded.

The documentation was incomplete because the setting impacts the request body encoding overall, not just when we are encoding a form, but also when we are multipart (e.g. uploads) or sending a single body (e.g. json).

I've clarified the doc: 3c36fd078bd9d694855c5fe8fae05e05d5c441ca

But the setting does not change the encoding of parameters in the request URL, and is not intended to. That always gets done into ASCII: https://github.com/jhy/jsoup/blob/6c0c2d4514278a82eb0868c5d978612a79c6d058/src/main/java/org/jsoup/helper/UrlBuilder.java#L23

That's per the HTTP RFC. I haven't seen issues regarding that.

JamesHertz:
The problem, is that the badly designed website I was interacting with expects form data on get requests and the I needed a way to tell `jsoup` to tell it to encode the query fields in a certain charset. I haven't looked at the project in some time, but that was the only way I found to do that.

If there is a way to do that now in `jsoup` I think that is good, if there isn't, I think it might be worth giving some thought to it.

PR #1903, opened 2023-02-23
Use Collection.isEmpty().

Isira-Seneviratne:


charlie-smma (approved):

jhy:
Sorry for the late reply on this one. I think I'm going to skip it, as I find the "not is even" a bit harder to parse than "size greater than zero".

OK to move it in to new code in the positive sense (when checking if the list is empty) but I don't think it needs a specific PR for the codebase, just incremental.

jhy:
(Actually I think I just don't have a strong preference to use either way - including not is empty - let's just take it as it comes in code)

PR #1907, opened 2023-02-27
Use Objects.requireNonNull().

Isira-Seneviratne:


jhy:
Thanks, but I don't see the value in this. We still use the Validate class (which I like), and the Validate class throws better exceptions than Objects.requireNonNull (as it trims the Validate class out of the exception stack trace).

https://github.com/jhy/jsoup/blob/c5df15f9bf05d3fac7bcd05ffd8943bbdccf406a/src/main/java/org/jsoup/helper/ValidationException.java#L18
https://github.com/chibenwa/jsoup/blob/c5df15f9bf05d3fac7bcd05ffd8943bbdccf406a/src/main/java/org/jsoup/helper/ValidationException.java#L18
https://github.com/Isira-Seneviratne/jsoup/blob/c5df15f9bf05d3fac7bcd05ffd8943bbdccf406a/src/main/java/org/jsoup/helper/ValidationException.java#L18

I really only see value in moving to more newly available Java class methods if they substantially clarify the intent of the code or meaningfully remove code duplication.

PR #1922, opened 2023-03-23
Replace with enhanced "for"

jeff87218:


jhy:
No thanks - I tend not to use the "enhanced for" in any potentially performance sensitive loop, as it allocates an extra Iterator object which requires extra GC.

PR #1944, opened 2023-04-20
Removed unnecessary constructor

seujorgenochurras:
List<Elements> extends Collection<Elements> 
this means that in that case they are literally the same

### DIFF
--- a/src/main/java/org/jsoup/select/Elements.java
+++ b/src/main/java/org/jsoup/select/Elements.java
@@ -36,10 +36,6 @@ public class Elements extends ArrayList<Element> {
         super(elements);
     }
     
-    public Elements(List<Element> elements) {
-        super(elements);
-    }
-    
     public Elements(Element... elements) {
     	super(Arrays.asList(elements));
     }


seujorgenochurras:
Ok, maybe it wasn't that unnecessary lol.
But why is it giving the error "Removed Constructor"?

jhy:
Because the super returns an `ArrayList<E>`, not an `Elements`, hence the override.

Please run the testcases prior to any further PRs.

seujorgenochurras:
> Because the super returns an `ArrayList<E>`, not an `Elements`, hence the override.
> 
> Please run the testcases prior to any further PRs.

Maven tests ran like butter.
I didn't understand, could you please clarify?

The super class is ArrayList, the constructor that I have removed utilizes the same super constructor already defined (the one that uses Collection instead)

jhy:
> Maven tests run like butter.
> I didn't understand, could you please clarify?

You should run `mvn verify` which will run all the verification plugins, including the japicmp plugin which checks backwards compatibility, and errors out here. In fact if you run that after making the change but without recompiling, you will see unit tests failing as well, with:

```
java.lang.NoSuchMethodError: 'void org.jsoup.select.Elements.<init>(java.util.List)'
```

> The super class is ArrayList, the constructor that I have removed utilizes the same super constructor already defined (the one that uses Collection instead)

It's source compatible but not binary compatible, as compiled classes will be looking for the List signature method.

The reason both methods are there in the first place is because early on, they had different implementations. The List version kept a local ref to the list, while the Collection version made a copy. That was simplified later. (You can see the history for more details if you're interested.)

seujorgenochurras:
Wow, thank you for that.
I really wasn't expecting a response.
Thanks for the clarification

PR #1945, opened 2023-04-20
Removed deprecated forEachNode method

seujorgenochurras:
The forEachNode method uses a Consumer FunctionalInterface which is literally the same as the one in JDK 8


jhy:
These can't be removed until a later release, so that there's a released version with the `@Deprecated` annotation published.

PR #1950, opened 2023-05-05
Add testcase and fix for attributes deduplication in form and empty elements

### LINKED ISSUE #1949: Attributes in form-elements are not deduplicated when parsing as HTML
I have an issue with form-elements that contains duplicate attributes. I was expecting that attributes in form-elements should be deduplicated in the same way as all other elements when parsing HTML via Parser.htmlParser() but duplicate attributes seems to always be retained. Using Parser.xmlParser(), form attributes are correctly deduplicated so the issue only affects HTML.

Looking though old issues and pull requests, I found #1219 that seems to fix deduplication for start tags but that fix doesn't seems to apply to form-elements.

Here is a simple test-case adopted from `HtmlParserTest`:

```
    @Test public void dropsDuplicateAttributesInFormElement() {
        String html = "<form One=One ONE=Two Two=two one=Three One=Four two=Five></form>";
        Parser parser = Parser.htmlParser().setTrackErrors(10);
        Document doc = parser.parseInput(html, "");

        Element p = doc.selectFirst("form");
        assertEquals("<form one=\"One\" two=\"two\"></form>", p.outerHtml()); // normalized names due to lower casing

        assertEquals(1, parser.getErrors().size());
        assertEquals("Dropped duplicate attribute(s) in tag [form]", parser.getErrors().get(0).getErrorMessage());
    }
```

perlan:
This is an attempt to handle deduplication of attributes in form-elements and should fix #1949

### DIFF
--- a/src/main/java/org/jsoup/parser/HtmlTreeBuilder.java
+++ b/src/main/java/org/jsoup/parser/HtmlTreeBuilder.java
@@ -228,7 +228,7 @@ public class HtmlTreeBuilder extends TreeBuilder {
     }
 
     Element insert(final Token.StartTag startTag) {
-        dedupliateTagAttributes(startTag);
+        dedupeAttributes(startTag);
 
         // handle empty unknown tags
         // when the spec expects an empty tag, will directly hit insertEmpty, so won't generate this fake end tag.
@@ -262,7 +262,7 @@ public class HtmlTreeBuilder extends TreeBuilder {
     }
 
     Element insertEmpty(Token.StartTag startTag) {
-        dedupliateTagAttributes(startTag);
+        dedupeAttributes(startTag);
 
         Tag tag = tagFor(startTag.name(), settings);
         Element el = new Element(tag, null, settings.normalizeAttributes(startTag.attributes));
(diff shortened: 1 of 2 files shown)


src/main/java/org/jsoup/parser/HtmlTreeBuilder.java
```diff
@@ -282,6 +282,14 @@ Element insertEmpty(Token.StartTag startTag) {
     }
 
     FormElement insertForm(Token.StartTag startTag, boolean onStack, boolean checkTemplateStack) {
+        // cleanup duplicate attributes:
```
jhy, line 285: I think that we should factor this out so the code is not duplicated on 230. Either a common method, or better, a shared code path.

jhy (changes requested):
Looks good, would like to get the duplicate code factored out.

jhy:
(BTW, I am happy to make that change myself if you prefer, LMK.)

perlan:
Thank you for responding to my PR. I took a quick look at the code path and I'm not familiar enough with the code to try any refactoring of the code path but I can certainly factor out the deduplication code.

perlan:
When moving the deduplication code to it's own method, I realized that deduplication was not applied for empty tags either so I added another test case and fix for empty tags as well.

src/main/java/org/jsoup/parser/HtmlTreeBuilder.java
perlan: Sorry. I already added comment below

jhy (approved):
Thanks, looks great. I added a commit that makes the test case parameterized, to remove some duplication.

jhy:
Thanks, merged!

PR #1957, opened 2023-05-18
Element - convenience methods for primitive type conversion.

sam-hieken:
Hello, I just threw in a couple of convenience methods for converting to primitive values from `text()`.


jhy:
Thanks for your interest, but I don't feel this is a good fit for jsoup at this time. These are single line methods and don't seem to add anything in the way of convenience over just calling the standard library. And if the user doesn't know the standard library, it's unlikely they'll know these jsoup specific methods.

PR #1964, opened 2023-06-03
Fix NPE in escaping if output settings cloned twice

TalgatAkhm:
Cloning OutputSettings leads to NPE during escaping
Code example:
```
    private static final Document.OutputSettings DEFAULT = new Document.OutputSettings();

    @Test
    public void test() {
        Entities.escape("text", DEFAULT.clone());
        Entities.escape("text", DEFAULT.clone());
    }
```

Stack: 
```
java.lang.NullPointerException: Cannot invoke "org.jsoup.nodes.Entities$CoreCharset.ordinal()" because "charset" is null

  at org.jsoup@1.15.3/org.jsoup.nodes.Entities.canEncode(Entities.java:298)
  at org.jsoup@1.15.3/org.jsoup.nodes.Entities.escape(Entities.java:239)
  at org.jsoup@1.15.3/org.jsoup.nodes.Entities.escape(Entities.java:145)
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,9 @@ Release 1.16.2 [PENDING]
   * Bugfix: `form` elements and empty elements (such as `img`) did not have their attributes de-duplicated.
     <https://github.com/jhy/jsoup/pull/1950>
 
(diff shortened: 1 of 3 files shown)


jhy:
Thanks, merged!

Issue #1428, opened 2020-09-23
Expose the method for registering new tags in XML and the possibility of saying "formatAsBlock" false or true


dcmarcelino:
In our system, we created some xml tags to facilitate our understanding, but some we want in the parser, to be considered as inline, because we need the number of characters of the text faithful to what was written, because without this, I was adding formatting spaces

exemple:
```
<assessment-group id="2C93808974B5C5DA0174B73EB5712FBA"> 
            <label> 2. </label> 
            <highlight show-answer="true" id="2C93808974B5C5DA0174B73EB5712FBB"> 
                <question> 
                    <p id="2C93808974B5C5DA0174B73EB5712FBC">Underline the best word to complete each sentence below.</p> 
                </question> 
                <items> 
                    <item id="2C93808974B5C5DA0174B73EB5712FBD">He has <token type="underline" color="any">received</token> many gifts, <strong>but</strong> his wife <strong>has</strong> received <strong>no /</strong> <token type="underline" color="any">none</token></item> 
                    <item id="2C93808974B5C5DA0174B73EB5712FBE">The <strong>bird</strong> hasn't drunk <token type="circle" color="any"><strong>any</strong></token> <strong> / some</strong> water today.</item> 
                    <item id="2C93808974B5C5DA0174B73EB5712FBF">Did you watch <strong>no / </strong> <token type="underline" color="any"> any </token> films last month?</item> 
                    <item id="2C93808974B5C5DA0174B73EB5712FC0">I have <token type="underline" color="any"> some </token> <strong> / any</strong> news for you.</item> 
                    <item id="2C93808974B5C5DA0174B73EB5712FC1">Steven is going to answer <token type="underline" color="any"> no </token> <strong>/ any</strong> e-mails tomorrow.</item> 
                    <item id="2C93808974B5C5DA0174B73EB5712FC2">There is <token type="underline" color="any"> no </token> <strong>/ any</strong> reason to worry about Dan.</item> 
                    <item id="2C93808974B5C5DA0174B73EB5712FC3">He has received many gifts, <strong>but</strong> his wife <strong>has</strong> received <strong>no /</strong> <token type="underline" color="any">none</token></item> 
                    <item id="2C93808974B5C5DA0174B73EB5712FC4">Would you care for <strong>any / </strong> <token type="underline" color="any"> some </token> finger food?</item> 
[...]

Issue #502, opened 2014-12-04
Allow configuration of self closing for tags


Artur-:
When using web components / custom elements the tag will always be marked as selfClosing=false and I can see no way of changing this

Legioth:
The parser already marks unknown (i.e. custom) tags as selfClosing if they are encountered as such in the input, but there is no such support for tags created programatically using Tag.valueOf.

I'm not sure whether it would make sense to just add an overload of valueOf that also takes a boolean for controlling the selfClosing field, or if it would be wiser to introduce a TagBuilder that would allow customising all aspects of the configurations affecting how the tag is output.

Legioth:
On the other hand, this might not make any sense at all since "html5" has no concept of self-closing tags as explained in e.g. http://stackoverflow.com/a/3558200 other than for "foreign" elements, i.e. MathML and SVG elements. The Custom Elements specification doesn't define such elements as "foreign" nor does it define any new syntax that would support self-closing tags.

gitastrophe:
Jsoup supports XML parsing and generation, not just HTML.  Support for non-HTML5 self-closing tags is mandatory for XML generation.  It would be much nicer to be able to either pass a flag to Tag#valueOf or modify the self-closing attribute.  I use the following utility method for creating self-closing elements in XML:

```
public static Element createVoidElement(String tagName, String baseUri) {

        Document document = Jsoup.parse("<" + tagName + "/>", baseUri, Parser.xmlParser());
        document.outputSettings().prettyPrint(false);
        document.outputSettings().escapeMode(Entities.EscapeMode.xhtml);

        return document.body() == null ? document.child(0) : document.body().child(0);
    }
```

Nopalin:
The problem allowing for change the self-close property is that a developer can set it as true and the element having children, which would result to an incorrect xml. I think the model is fine, just add some static methods for allowing us to register our tags, because no body has dinamic tags, we all has specific tags for our project, that can be defined in developing time. If the needs increases, the tags increases or changes, and that's it.

soorapadman:
Is there a way i can create  a tag **<my-tag/>**  like this using jsoup?

nicoloboschi:
For who needs to convert self closing tags to non-self closing tags, there is a tricky solution using PowerMock

```
 Document document = Jsoup.parse(...);
document.traverse(new NodeVisitor() {
            @Override
            public void head(Node node, int depth) {
                if (node instanceof Element) {
                    Element el = (Element) node;
[...]

PR #1980, opened 2023-07-13
Add method to rename an Element's Tag without resetting the tag's properties

ThoughtProcess:
This PR is tangentially related to #1428 and #502.

If I want to change the name of an element's tag as I traverse the DOM, I don't have the ability to "preserve" the properties of the underlying Tag object (for example, keeping a tag as self-closing).  A great example of this is renaming a JSP tag, where I might want to change a `<bean:message />` to a `<spring:message />`.  

This PR adds a new method to the Element class, `renameTagPreserveProperties(String)`, to support that behavior.  Also added unit tests to confirm that the behavior of the `Element.tagName(String)` method is preserved.

### DIFF
--- a/src/main/java/org/jsoup/nodes/Element.java
+++ b/src/main/java/org/jsoup/nodes/Element.java
@@ -178,6 +178,20 @@ public class Element extends Node {
     }
 
     /**
+     * Change (rename) the tag of this element. For example, convert a {@code <span>} to a {@code <div>} with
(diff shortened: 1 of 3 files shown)


jhy:
OK implemented setters now for name, namespace, and options set() and clear(). On the [develop branch](https://github.com/jhy/jsoup/compare/master...develop-1.20.1).

PR #1988, opened 2023-07-29
Uncomment stream block in Element

Isira-Seneviratne:
Uncomment the `stream` block in `Element` and make some improvements to it.


jhy:
While this is a bit neater, I prefer the current version because it recycles in-use StringBuilders, vs allocating a new one every time. As this method gets called via `Element#cssSelector()` in a loop, that can be a pretty impactful cause of garbage objects.

Perhaps modify the StringUtil to implement a Collectors.joining that uses recycled builders?

src/main/java/org/jsoup/internal/StringUtil.java
```diff
@@ -5,11 +5,14 @@
 
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Iterator;
```
jhy, line 10: Can you change your format settings so that we don't use wildcard imports? Prefer to list them all explicitly.

jhy:
Thanks, merged!

PR #1989, opened 2023-07-29
Switch to Java's UncheckedIOException.

Isira-Seneviratne:


jhy:
This is failing because we currently have

```xml
<excludes>
  <exclude>@java.lang.Deprecated</exclude>
</excludes>
```

configured in japicmp. This allows the previously deprecated methods (between 1.16.1 and removed in 1.16.2) to not blow up. But I can't see a way to configure japicmp to allow this to be marked deprecated while that exclude is on -- it makes the class invisible and therefore looks like a removal. So I think we need to pick it up after the next release.

Also I think we should change the signature to

```java
public class UncheckedIOException extends java.io.UncheckedIOException 
```

(and throw the extended not base version) for a release. That way if anyone is explicitly catching this exception, the current code will still work and they have an opportunity to migrate.

PR #1990, opened 2023-07-29
Use computeIfAbsent()

Isira-Seneviratne:


src/main/java/org/jsoup/select/StructuralEvaluator.java
```diff
@@ -24,18 +25,9 @@ public StructuralEvaluator(Evaluator evaluator) {
...
-            memo = new IdentityHashMap<>();
-            rootMemo.put(root, memo);
-        }
-        Boolean matches = memo.get(element);
-        if (matches == null) {
-            matches = evaluator.matches(root, element);
-            memo.put(element, matches);
-        }
-        return matches;
+        Map<Element, IdentityHashMap<Element, Boolean>> rootMemo = threadMemo.get();
```
jhy, line 28: Please see the code comment in line 27 -- I tried this before and under profiling, there is a ton of garbage generation for the lambda closure; a new object on every method call.

Perhaps make the supplier a static class vs a lambda?

src/main/java/org/jsoup/helper/HttpConnection.java
```diff
@@ -1096,9 +1096,9 @@ private Response(HttpURLConnection conn, HttpConnection.Request request, HttpCon
             }
         }
 
-        private static LinkedHashMap<String, List<String>> createHeaderMap(HttpURLConnection conn) {
```
jhy, line 1099: I prefer this returning a LinkedHashMap to be clear its an ordered map.

jhy (changes requested):
Thanks -- added some comments inline

jhy (approved):

jhy:
Thanks, merged!

PR #2004, opened 2023-09-25
An Annotation-Based Way to Convert From `Element` / `Elements` to Java objects (Jackson/GSON style).

sam-hieken:
Hello,

I've noticed there's no built in way to turn XML / HTML into plain Java objects, and decided to build out a solution for my personal use. It's been a massive productivity boost (especially when dealing with XML), so I figured it couldn't hurt to see if anyone was interested in getting it merged as an experimental feature.

- [Demonstration](https://pastebin.com/2XwwXerh)
- [Example data (needed for demo)](https://pastebin.com/Y5bgiRjK)

Please keep in mind a few things:
- I started working on this yesterday; while I have done quite a lot of testing, I wouldn't be surprised if there are some issues in terms of error handling and performance.
- Conversion is currently only _forward_: I currently haven't added a way to convert _back_ from a properly annotated object to an `Element` / `Elements` (nor do I plan to at the moment)
- The only supported collection types currently are `List` and `Set` for simplicity, though I'd be open to adding support for most (if not all)
- Documentation is sparse; due to how recently I created this (and the fact I created it for personal use), documentation is not my priority (more than happy to build it out though if this gets merged)
- My testing has mainly been parsing XML: it's possible some issues will arise using HTML (though I can't think of any HTML specific issues, as the converter should work with all SGML).

Thank you


jhy:
Hi there -- this looks quite interesting, but I don't feel it's a good fit directly in jsoup. I think it would be better as a separate specific project, leveraging jsoup for the parsing. On first review I believe no modifications to jsoup are required; please ping me if otherwise.

Cheers,
Jonthan

Issue #2009, opened 2023-10-11
formData retrieve "input type=image" incorrectly


ericksonlbs:
Method "formData()" from FormElement retrieve "input type=image" incorrectly

In the HTML form, in the submission form "input type=image" is not sent in the body, but using JSoap it is retrieved.

PR #2010, opened 2023-10-11
formData retrieves "input type=image" incorrectly

ericksonlbs:
This test should pass when fix issue #2009


jhy:
Looks good, thanks!

PR #2019, opened 2023-10-25
readd an implementation of Whitelist for backwards compatibility

bangert:
First off, thanks for jsoup - i've used it in multiple settings and it is a pleasure to use.

In this case however, we are not the direct user of jsoup. We use some software, where we do not control the dependency chain and that still depends on jsoup <1.15.x. For security reason we would like to update to the latest version - as is also the recommendation in the security policy[1] . However the recent ABI breaking change renaming the Whitelist class to Safelist prevents us from doing so.

This PR readds a Whitelist class, that uses a Safelist internally. It works for us and it may be useful to others.
Thanks

[1] https://github.com/jhy/jsoup/blob/master/SECURITY.md

### DIFF
--- a/src/main/java/org/jsoup/Jsoup.java
+++ b/src/main/java/org/jsoup/Jsoup.java
@@ -365,18 +365,22 @@ Connection con3 = session.newRequest();
         return new Cleaner(safelist).isValidBodyHtml(bodyHtml);
     }
 
+    @Deprecated
     public static String clean(String bodyHtml, String baseUri, Whitelist whitelist, Document.OutputSettings outputSettings) {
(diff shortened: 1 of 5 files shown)


jhy:
Hi there - thanks for your feedback on jsoup, I'm glad to hear its been useful for you!

And thanks for the code sample. It's similar to the compatibility shim (#1464) we provided between `1.14.1` (2021-Jul-10) where the Whitelist was deprecated in favor of Safelist, and `1.15.1` (2022-May-15) when it was ultimately removed.

bangert:
ah - i didnt know that. less than a year is pretty ambitious, but i understand that it is difficult to get downstreams to upgrade if they dont have too - which this PR is a testament of, since they sometimes dont even upgrade even though they have to...

PR #2024, opened 2023-10-30
Java 21 CI build

offa:
Adds a CI build for Java 21 (LTS) and some Action updates.


jhy:
Thanks! Merged. I also removed JDK 20 as it seems unnecessary now. When 22 comes out we may remove 17.

Issue #1992, opened 2023-08-24
FR: Strict Nullability? 


stefanhaustein:
Hi, 

Unfortunately, the current nullability annotations don't work in all the places where they are needed for strict null-marking. For instance, we currently can't correctly mark the generic value type of the Attribute Map Entry as nullable with javax.annotations.Nullable:

```
public class Attribute implements Map.Entry<String, @Nullable String>, Cloneable  {
```

Would you mind if I'd create a PR to switch the project to strict nullability using [Jspecify](https://github.com/jspecify/jspecify/wiki/adoption)? This would also allow us to switch the two internal "NonNullByDefault" annotations over to @NullMarked.

Stefan

PR #2025, opened 2023-11-01
Repackaged the library with native (vs automatic) JPMS module support

### LINKED ISSUE #1466: Provide module-info.java for Java9+ users
I know you already provide an Automatic-Module-Name for Java9+ users but we would be better off with a multi-release JAR that exports a full-fledged `module-info.java` because when using `requires transitive org.jsoup` the compiler complains (warns) against exporting automatic modules.

Publishing a MultiRelease JAR using Maven should be relatively easy. See `multiReleaseOutput` at https://www.baeldung.com/maven-multi-release-jars

Hit me up if you run into any problems and I'll take a look.

### LINKED ISSUE #1943: Create a java module version of jsoup-xxx.jar
I am using jsoup in a javafx desktop app project and jsoup is working beautifully (thank you). However, the jsoup-xxx.jar file is not a java module. It is taken as an automatic module while building the project with module-info.java and everything works fine. It breaks when I try to create a runtime image with jlink. I suspect it is because jsoup jar is not a java module. Is it possible to make the library available as a java module?

jhy:
Adds an explicit `module-info.java` file. As a native module, tools requiring modules such as jlink will operate correctly without warnings.

This makes jsoup a multi-release jar so it's still compatible with java 8. The `module-info.java` file is:

```java
module org.jsoup {
    exports org.jsoup;
    exports org.jsoup.helper;
    exports org.jsoup.nodes;
    exports org.jsoup.parser;
    exports org.jsoup.safety;
    exports org.jsoup.select;

    requires transitive java.xml; // for org.w3c.dom out of W3CDom
}
```

**Nullability annotations**
I looked at having also a `requires static jsr305` entry as well. That's a build time dependency for our javax nonnull annotations. However as that's an automodule, we get the compile warning `Required filename-based automodules detected: [jsr305-3.0.2.jar]. Please don't publish this project to a public artifact repository!`. So I'm keeping it out. Everything still works and downstream consumers will still get nullability warnings in their IDE. However jsoup developers will see edit-time errors / warnings on the import and use of these annotations. Perhaps now is a good time to implement #1992.

**Java 8 tests**
This change requires a minimum Java version of 9, and the min CI builder is changed to 11 (LTS). We should add a CI action to test a built jar under Java 8. Tracked in #2026.

Fixes #1943
Fixes #1466

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         # choosing to run a reduced set of LTS, current, and next, to balance coverage and execution time
-        java: [9, 17, 21]
+        java: [11, 17, 21]
       fail-fast: false
     name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         # choosing to run a reduced set of LTS, current, and next, to balance coverage and execution time
-        java: [8, 17, 21]
+        java: [9, 17, 21]
       fail-fast: false
     name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}
     steps:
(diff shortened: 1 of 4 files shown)


jhy:
Thanks @cowwoc, that's helpful. I'm tracking the nullability assertion change in #2028. I recall that when we chose the javax package vs just local names it was to ensure Kotlin understood the assertions. But in practice I don't know that that does much.

PR #2027, opened 2023-11-01
Move the multi-release to a JDK9+ profile

### LINKED ISSUE #2026: Add a JRE 8 CI test action
To implement JPMS module support(#2025), the minimum JDK version for building jsoup becomes JDK 9. However we still will run on JRE 8. We do have the Animal Sniffer plugin to validate JRE 8 compatibility, but it would be better if we also had a specific CI action to test.

So this task is to add an action that executes after the build action (which will need to change to save the built jar into cache) and then runs the tests on JDK 8.

jhy:
Allows JDK8 to build, albeit without multi-release / modules. Allows the CI actions to test on Java 8.

Fixes #2026

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         # choosing to run a reduced set of LTS, current, and next, to balance coverage and execution time
-        java: [11, 17, 21]
+        java: [8, 17, 21]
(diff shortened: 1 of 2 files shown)


jhy:
Validated in the CI builds that Java 8 succeeds (and does not include module-info); newer versions succeed and do include. In newer versions, source/target for all but module-info is 1.8 and so jars built on those are compatible in Java 8.

jhy:
Need to fix up nullable annotations, see #2028

PR #2048, opened 2023-11-10
Migrate Nullability assertions to jspecify

### LINKED ISSUE #1992: FR: Strict Nullability?

### LINKED ISSUE #2028: Migrate to module-supported Nullable annotation package
To mitigate errors/warnings in IntelliJ (but not present in the real Maven build), I'm temporarily adding `requires static jsr305;` to `module-info.java`, so that the nullable annotations are visible. But, we don't want to produce a published library with this as it's an automodule.

Need to select another Nullable package (#1992 jspecify?) before next release. Something that IDEs and Kotlin and other various tools will recognize.

jhy:
Away from jsr305

With the goal of having a nullability assertion annotation that supports Java modules, so we can release the next version of jsoup.

Annotation options appear limited and does not support @​WillClose, but rest appears OK.

Fixes #2028 
Fixes #1992

Related: #1616

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,7 +173,7 @@
           <instructions>
             <Bundle-DocURL>https://jsoup.org/</Bundle-DocURL>
             <Export-Package>org.jsoup.*</Export-Package>
-            <Import-Package>javax.annotation;version=!;resolution:=optional,javax.annotation.meta;version=!;resolution:=optional,*</Import-Package>
+            <Import-Package>org.jspecify.annotations;version=!;resolution:=optional,*</Import-Package>
           </instructions>
         </configuration>
       </plugin>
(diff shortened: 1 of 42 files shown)


jhy:
For reference (as noted in #2028):

To get jspecify nullability warnings to show in IntelliJ:

Settings -> Editor -> Inspections -> Probable bugs -> Nullability Problems -> @​NotNull/ @​Nullable problems:
Configure Annotations
Add `org.jspecify.annotations.Nullable` to the Nullable tab. 
And `org.jspecify.annotations.NullMarked` and `org.jspecify.annotations.NonNull` to the Not Null tab.

PR #2055, opened 2023-11-19
Add Path-accepting Jsoup methods

Isira-Seneviratne:
* Add Jsoup methods that accept a [`Path`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html).
* Also improve file parsing using [`SeekableByteChannel`](https://docs.oracle.com/javase/8/docs/api/index.html?java/nio/channels/SeekableByteChannel.html) by resetting the current position after reading the magic bytes of the file being parsed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # jsoup Changelog
 
+## 1.18.1 (Pending)
+
+* Added `Path` accepting parse methods: `Jsoup.parse(Path)`, `Jsoup.parse(path, charsetName, baseUri, parser)`,
+  etc. [2055](https://github.com/jhy/jsoup/pull/2055)
+
 ## 1.17.2 (Pending)
 
 ### Improvements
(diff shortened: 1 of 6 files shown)


Isira-Seneviratne:
> Some tests using the API would be a good addition.
> 
> > * Also improve file parsing using [`SeekableByteChannel`](https://docs.oracle.com/javase/8/docs/api/index.html?java/nio/channels/SeekableByteChannel.html).
> 
> Could you expand on that, in what ways does using a channel here improve the parse?

`SeekableByteChannel` allows the current position to be changed, so I used that functionality instead of closing and reopening the input stream: https://github.com/jhy/jsoup/pull/2055/files#diff-decf16c38aafa83f24bc30e25981eced9551a4feb5ab5af5238f74c3ad839b1fR123

jhy:
Thanks! Merged

PR #2062, opened 2023-11-26
Refactor: Fix "Feature envy" design level smell in the class "Cleaner"

smit0086:
This PR attempts to fix the "Feature envy" smell detected in the class "Cleaner" using the tool [Designite](https://www.designite-tools.com).

**Description of the smell**
The tool detected a instance of this smell because createSafeElement is more interested in members of the type: Safelist

**How I fixed the smell?**
To fix the code smell, I have moved the method createSafeElement from the Cleaner class to the class Safelist, and updated it's usage inside the Cleaner class.


jhy:
This just seems like change for change's sake. Current impl is fine. And this needlessly elevates and exposes createSafeElement and ElementMeta to the contractual API.

PR #2063, opened 2023-11-27
Refactor: Fix multifaceted abstraction design level smell

smit0086:
The class DataUtil has multiple responsibilities, In this PR I have created another class CharsetUtil to extract Charset related utilities from the DataUtil class.


jhy:
This just seems like change for change's sake. I prefer the current version of a larger class - easier to see the implementation of each method. Plus, I don't want the internal method validateCharset to be made public (and not documented).

PR #2092, opened 2023-12-28
Add select methods returning element streams

Isira-Seneviratne:


jhy (changes requested):
Looks cool, thanks! Can you see the inline comments.

And, can you add tests. Would be good to create a testcase that compares some queries with the result of a `select(query).stream()`.

jhy:
I liked the original PR but I don't think I'm a fan of the changes in e1e35bb627ae25130e2b7df505580061f6bf8284 - all the getElementsStream methods. Generally the getElementBy... methods are only there to help users get used to jsoup coming from APIs like the W3C DOM APIs. They are not as powerful as the select methods. And extending them for direct stream access feels like its adding a bit of clutter for little value.

I think the selectStream(query) and selectStream(evaluator) should cover the core use case of this.

Also I think the test is still pretty anemic as it only selects one element. Doesn't test conditions where there's more or less results, or the order, etc. Would be great if you could beef those up.

Isira-Seneviratne:
Fair enough. I was planning to add more tests, should've marked this PR as draft.

jhy (approved):
Looks good, thanks! I tweaked the docs a bit and added a simple test to verify the order (which was already implicit because the existing Elements Selector tests effectively do that anyway)

PR #2095, opened 2023-12-30
Improve equals and hashCode implementations using Objects

Isira-Seneviratne:


jhy:
Thanks, merged.

One point - the varargs method `Objects.hash(Object... values)` can create a lot of short-lived GC pressure when called as it creates wrapping Arrays for each invocation. I think in this case it is OK because the code looks so much better, and these hashcodes aren't typically called in performance critical loops. But I have removed other instances of varargs e.g. in the TreeBuilder after finding perf impacts. So just something to keep in mind for other code areas.

PR #2099, opened 2024-01-09
minor optimization

821938089:
Benchmarks are not significantly improved, maybe about 1% faster as observed by profiler (tested on Android platform).


jhy:
Thanks, merged. Though would ask in the future to provide the benchmarks so we can review.

PR #2113, opened 2024-01-30
Use TreeMap to store headers

Isira-Seneviratne:
Use the `TreeMap` class to store the headers, as it allows keys to be checked case-insensitively.


jhy:
Hmm - this looks neater, but it changes the semantics. The LinkedHashMap implementation keeps the headers in the order they are added. But the TreeMap will sort them alphabetically. The LinkedHashMap is a deliberate choice to maintain the original order.

I thought we had tests to verify the order was maintained but I guess not. That would be good to add...

jhy:
@Isira-Seneviratne I think we need to close this one, as mentioned, using a TreeMap will lose the insert order of the headers. I want to keep the insertion order so that it reflects the order that they were sent by the server.

LMK if you have another idea for the case insensitive scan.

PR #2114, opened 2024-02-01
Fix NullPointerException in case a custom document is passed to W3CDom

casid:
When we provide a custom output `Document` to `W3CDom#convert`, we receive a `NullPointerException` in the `W3CBuilder` constructor.

I've added a test that demonstrates the behavior. Here's the stacktrace from that test without the applied fix.

```
java.lang.NullPointerException: Cannot invoke "org.jsoup.nodes.Element.ownerDocument()" because "this.contextElement" is null

	at org.jsoup.helper.W3CDom$W3CBuilder.<init>(W3CDom.java:364)
	at org.jsoup.helper.W3CDom.convert(W3CDom.java:250)
	at org.jsoup.helper.W3CDom.convert(W3CDom.java:238)
	at org.jsoup.helper.W3CDomTest.canConvertToCustomDocument(W3CDomTest.java:199)
```

### DIFF
--- a/src/main/java/org/jsoup/helper/W3CDom.java
+++ b/src/main/java/org/jsoup/helper/W3CDom.java
@@ -360,12 +360,14 @@ public class W3CDom {
             namespacesStack.push(new HashMap<>());
             dest = doc;
             contextElement = (org.jsoup.nodes.Element) doc.getUserData(ContextProperty); // Track the context jsoup Element, so we can save the corresponding w3c element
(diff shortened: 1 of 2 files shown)


casid:
Hi @jhy,

did you get a chance to have a look at this already?

It is not urgent, but it prevents us from upgrading to the latest JSOUP version without hacking the classpass.

Thank you so much!

jhy:
Great, TY!

casid:
Thank you @jhy!

PR #2119, opened 2024-02-15
Add uses of StringUtil.joining()

Isira-Seneviratne:


jhy (approved):
Thanks, looks good. I made some tweaks to the format.

PR #2124, opened 2024-02-28
Adding test cases to improve test coverage and refactoring to clean up design and implementation code smell

bhishman-desai:
Adding test cases for Range, FormElement and Attribute class to improve overall coverage.


jhy:
Thanks, but I don't think I'll merge this. This is changing various components at once, and I don't really agree with all of the changes, or see any particular improvement.

PR #2151, opened 2024-07-01
Use zulu vs temurin

jhy:
For JDK 8+ support, incl on ARM


jhy:
Fixes build failures on GitHub MacOS ARM runners. See https://github.com/actions/setup-java/issues/625

PR #2173, opened 2024-07-16
Bump minimum Android API level to 21

Isira-Seneviratne:
Android versions below 5.0 (API level 21) are obsolete these days. This change would allow the use of features such as `StandardCharsets`, which is available on Android 4.4 (API level 19) and higher.

This PR also changes one Sniffer dependency to [Gummy Bears](https://github.com/open-toast/gummy-bears), which has support for desugaring.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,13 @@
 
 ## 1.19.1 (PENDING)
 
-### Bug Fixex
+### Changes
+
+* Updated the minimum Android API Level validation from 10 to **21**. As with previous jsoup versions, Android
+  developers need to enable core library desugaring. The minimum Java version remains Java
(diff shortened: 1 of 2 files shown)


jhy:
Is there anything other that might be useful other than StandardCharsets? That doesn't really make any difference from my POV, being that we statically initialize UTF-8 anyway at startup.

I get that the usage is very low % wise, but I don't see a reason to remove support just because?

Isira-Seneviratne:
There's also try-with-resources.

jhy:
OK, try-with-resources will be good to have. I have marked this for the 1.19.1 milestone (vs 1.18.2), as I feel that the change warrants being in a major bump.

PR #2179, opened 2024-07-23
Use Files.list() in fuzz tests

Isira-Seneviratne:
Use the [`Files.list()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#list-java.nio.file.Path-) method in the fuzz tests.


jhy:
Thanks, but I don't see any real benefit here? It's just, different.

And, we lose the assertion that we actually found any fuzz tests to perform.

PR #2183, opened 2024-07-30
Simplified the Entities.escape method

jhy:
Introduced an options bitset instead of all those boolean method options. Reduced cyclomatic complexity of from 29 to 14.

Improved throughput of escape around 22% if the content contains characters in supplemental plane, by no longer going through the charset encoder to test can encode, but pushing it into the CoreCharset. That removes a ByteBuffer allocation on each hit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@
 * In the `TreeBuilder`, the `onNodeInserted()` and `onNodeClosed()` events are now also fired for the outermost /
   root `Document` node. This enables source position tracking on the Document node (which was previously unset). And
   it also enables the node traversor to see the outer Document node. [2182](https://github.com/jhy/jsoup/pull/2182)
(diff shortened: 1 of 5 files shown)


jhy:
I perf tested using this wiki [list of colors](https://en.wikipedia.org/wiki/List_of_colors_(alphabetical)), which has some interesting attributes like `<p title="𝗛𝗦𝗩 (34° 14% 98%)&#10;𝗥𝗚𝗕 (250 235 215)&#10;𝗛𝗘𝗫 #FAEBD7">`. Those are `mathematical sans-serif bold capital h` etc; `&#120283;` if encoded in ASCII (in UTF we don't need to encode).

Average `html()` operations per second wen from ~ 156 to ~ 186.

jhy:
And some further improvements in 27e7d5f03d85bea15fbfe5432c37c6de94965a82. Getting around 235 ops/sec (from 156 -- so about 49% faster now!)

PR #2217, opened 2024-10-19
Use Map#getOrDefault() method

Isira-Seneviratne:
Use [`Map#getOrDefault()`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#getOrDefault-java.lang.Object-V-) in the `Safelist` class.


jhy:
Thanks but I don't really see that this particularly simplifies or makes it easier (for me!) to read.

Planning on overhauling the cleaner in #2284 so perhaps we can revisit then.

PR #2225, opened 2024-11-16
[JSOUP-2224] add wildcards

beargiles:
Added wildcards attributes for both tags and global scope. This allows the user to specify an unknown set of attributes, e.g., `data-*`, using a wildcard pattern. This is different from `:all` since it is selective.

This pattern must be recognized by `java.util.regex.Pattern`, e.g., `data-.+`. The pattern is case-insensiitve and must match the entire string since '^' and '$' are quietly added.

A variant approach is to extend the existing `addAttributes()` and `removeAttributes()` methods to recognize patterns, e.g., via a marker like enclosing braces. (hence `{data-.+}`). This approach could use the same implementation as below - the existing methods would simply strip the braces and call the now-private methods.

### DIFF
--- a/src/main/java/org/jsoup/safety/Safelist.java
+++ b/src/main/java/org/jsoup/safety/Safelist.java
@@ -63,13 +63,14 @@ import static org.jsoup.internal.Normalizer.lowerCase;
  */
 public class Safelist {
     private static final String All = ":all";
+    private static final TagName AllTagName = new TagName(All);
 
     private final Set<TagName> tagNames; // tags allowed, lower case. e.g. [p, br, span]
     private final Map<TagName, Set<AttributeKey>> attributes; // tag -> attribute[]. allowed attributes [href] for a tag.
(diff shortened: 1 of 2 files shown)


beargiles:
I'll check the error...

After thinking about it overnight I think it makes sense to remove the explicit 'wildcard' methods and use `pattern:...` for the attribute name. This is motivated in part by identifying a few other places where wildcards make sense (and thus would result in a combinatorical explosion if there are extensions, e.g., for more advanced pattern matching) and in part by having more flexibility in how the pattern is specified. There's no need for different braces/brackets/etc, just use a different protocol than 'pattern'.

I also realized that I need to add corresponding functionality to the add/remove protocol methods.

On a related note the 'All' value should be exposed. That would be used for global wildcards.

beargiles:
I almost forgot - the All should probably be extended with FormControl, Media, and possibly Visible. These would be used with the standard event attributes - it's shorthand for multiple related tags.

jhy:
Thanks -- I plan on overhauling the Cleaner and Safelist interfaces. See #2284. So I won't merge as-is, but will certainly re-review as that comes together.

PR #2228, opened 2024-11-16
[JSOUP-2227] Added HTML5 tags

beargiles:
Added HTML5 tags.

This is currently POC. The biggest question is that I've included all of the documented attributes but I've noticed that this is not done with the other tags.  I'm not sure why - it means there's a lot more work if you want to use 'clean' to strip out non-standard attributes but leave the rest.


jhy:
Thanks - won't merge as-is but I do plan on overhauling the Cleaner and Safelist as defined in #2284. Will see what we can pull in.

PR #2238, opened 2024-11-29
Introduce ErrorTracker class for managing error limits

hirvaDhandhukia:


jhy:
Hi,

This isn't a meaningful change but just looks like code churn and unnecessary change for the sake of change. And the code is not production ready.

Please don't submit further PRs like this.

Thanks!

PR #2257, opened 2025-01-07
Support http/2 requests via HttpClient

jhy:
This adds support for executing HTTP requests using the Java 11+ HttpClient on systems that support it, enabling **http/2** requests. On Java 8, and on Android, requests will still go via the existing HttpURLConnection.

This also gives a path for http/3 support when JEP [HTTP/3 for the HTTP Client API](https://openjdk.org/jeps/8291976) lands in Java.

This is currently disabled by default -- set the system property `jsoup.useHttpClient` to enable it. Once it gets tested and used by folks for a while, I plan to make it the default (for 1.20.x). Please do test it and let me know of any issues. It should largely be a drop-in replacement for the existing behavior.

This is implemented using the multi-release JAR feature, where a `HttpClientExecutor` is provided in the `/java11` version directory. When executed on earlier Java versions, that code is not found and the compatible `UrlConnectionExecutor` is used instead.

A couple notes:
* I moved the existing `module-info.java` from Java 9 to Java 11, so that we only need one module definition. That allows simple builds using IntelliJ (for e.g.) to continue to work, without requiring the full Maven build on each iteration. Given that 9 is not a LTS release, I don't expect that change to be an issue.
* For future work, would like to look at moving the HttpClient initiation from per-request to per-session; this may streamline repeated connection requests to the same hosts.
* All the 67 existing tests in `ConnectTest` are run both with HttpURLConnection and the new HttpClient impl and pass. As do the Session tests, and integration tests -- with the exception of the ability to interrupt the execution via a Thread.interrupt (as it executes in a managed worker thread; but timeouts are reliably and a better way to stop execution).


phothiwintxphl53-hue:
Powim

PR #2263, opened 2025-01-21
Add `selectFirst` and `expectFirst` to Elements

### LINKED ISSUE #2256: FR: selectFirst() and expectFirst() on org.jsoup.select.Elements
It'd be useful to be able to use `selectFirst()` and `expectFirst()` on `org.jsoup.select.Elements` instances to mirror the methods in `org.jsoup.nodes.Element`.
As far as I understand this is because `org.jsoup.select.Selector` provides both `select(String, Element)` and `select(String, Iterable<Element>)` but only `selectFirst(String, Element)`.

I could take a look at it myself, but people knowledgeable about jsoup internals would probably take much less time.

Bonus points if it's actually faster than `select(...).first()` 😉

Thanks for your work

midgleyc:
Closes #2256.

I copied the docstrings, and behaviour, from the equivalent methods on `Element` and the surrounding methods on `Elements`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,8 @@
   planner. [2254](https://github.com/jhy/jsoup/issues/2254)
 * Removed the legacy parsing rules for `<isindex>` tags, which would autovivify a `form` element with labels. This is no
   longer in the spec.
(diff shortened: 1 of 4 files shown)


jhy:
Cool, merged, thanks! I tweaked the formatting just to align with current style (which does diverge a bit from the existing code), and added a test to validate the root works the same as when selecting from an element.

PR #2283, opened 2025-03-09
Restore the behavior of cssSelector before v1.16.2 for elements with an id

### LINKED ISSUE #2282: `cssSelector` behavior changed since v1.16.2
Before v1.16.2, calling `cssSelector()` on the "A" paragraph in
```
<div>
  <div id="id1">
    <p>A</p>
  </div>
</div>
```
produced the result `#id1 > p`.

From v1.16.2, the result becomes `html > body > div > div > p` which is bound more to the structure of the HTML, and doesn't utilize the supposedly unique id anymore.

The [cssSelector doc](https://jsoup.org/apidocs/org/jsoup/nodes/Element.html#cssSelector()) also says
> If the element has an ID, returns #id; otherwise returns the parent (if any) CSS selector, followed by '>', followed by a unique selector for the element (tag.class.class:nth-child(n)).

which very likely leads to the understanding that the higher precedence of ID also applies to the parent CSS selector.

PhuNH:
Fixes #2282 

- Utilize id for not only the element itself, but also for its parent elements;
- Stop the processing when encountering an element with a unique id.

In case of an extremely deeply nested element, which is the reason for the changes in v1.16.2, this can possibly greatly reduce the size of the output of `cssSelector` if one of the element's parents has an id.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # jsoup Changelog
 
+## 1.19.2 (PENDING)
+
+### Changes
+
+### Improvements
+* `Element.cssSelector()` will prefer to return shorter selectors by using ancestor IDs when available and unique. E.g. `#id > div > p` instead of  `html > body > div > div > p` [#2283](https://github.com/jhy/jsoup/pull/2283).
+
 ## 1.19.1 (2025-03-04)
 
 ### Changes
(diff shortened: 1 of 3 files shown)


jhy:
Thanks! Merged. I tweaked the impl a little to not have to call id() and ownerDoc() as much.

PR #2292, opened 2025-03-28
Refactor Implementation and Design Smells

2612-Parva:
Complex conditional Use Decompose conditional, Extract method, Introduce Explaining Variable in Attribute.java
 Deficient encapsulation by using move method, extract class, replace conditional with polymorphism in ListLink.java
 Complex method by using Extract Method in AuthenticationHandler.java
 Complex Method and Long Statement Decompose conditional and Introduce explaining variable into DataUtil.java
 Unnecassry Abstraction by using Pull down method in StringUtil.java


jhy:
You've got to be kidding me...

PR #2308, opened 2025-04-21
ignoring unnecessarily generated surefire report

Talank:
In our analysis, we observe that this target/surefire-report/ directory is generated but not later used during the CI build. Hence, we propose to disable generating this directory to save runtime.
The generation of this directory can be disabled by simply adding -DdisableXmlReport=true to the mvn command.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,5 +28,5 @@ jobs:
         run: mvn -X compile -B --file pom.xml
 
       - name: Maven Verify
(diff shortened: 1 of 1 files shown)


jhy:
I don't see any particular value from worrying about that.

In fact it might be nice to upload all those generated artefacts somewhere. It would be helpful to have those reports when debugging a failure, as grepping through the Maven log in the GitHub truncating output is kind of a PITA.

PR #2312, opened 2025-04-24
Only match ASCII letters and digits

cketti:
Replace calls to `java.lang.Character.isLetter()` and `isDigit()`. In all these cases we really only want to allow ASCII letters or digits.


jhy:
Cool beans. Could you add a couple positive / negative tests for these, in StringUtils? So that nobody gets a rush of blood to their head and tries to tidy it up back to Character.isLetter.

cketti:
Done.

PR #2318, opened 2025-05-01
Add SimpleStreamReader

jhy:
Continuing on the work in #2186 to reuse large buffers to minimize allocations (and therefore GC) during parsing, this change allows us to reuse the byte -> char decoding buffer.

Replaces uses of InputStreamReader with a simplified version, where we recycle the underlying byte buffer, vs creating a new 8k buffer on every parse.

Gives a good performance improvements for all but the largest inputs, where the allocation cost is amortized across a longer parse.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
   <groupId>org.jsoup</groupId>
   <artifactId>jsoup</artifactId>
-  <version>1.20.2-SNAPSHOT</version><!-- remember to update previous version below for japicmp -->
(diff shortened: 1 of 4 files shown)


jhy:
benchmark

Here's a chart of the impact. The specific ops/s change is quite dependent on active memory pressure. Interesting to see that the small size shows little impact relative to the others. Not sure if that's a measurement anomaly of the test or if there's further ground to be made particularly around that size.

PR #2324, opened 2025-05-15
Selectors can match against Nodes

jhy:
This extends the QueryParser and Evaluators to support matching nodes (including specific node types, and their nodeValue contents)

Allows finding e.g. Elements following a particular comment or text node.

Also allows selecting nodes directly via 
- `Element#selectNodes(String css)` and 
- `Element#selectNodes(String css, Class nodeType)`

E.g.: `::comment:contains(prices) + p` will find `p` Elements directly after a `<!-- prices: -->` comment.

Or: `div:has(>::comment)` will find `div` Elements that directly have comment children.

Available operators:

<table>

 <tr><td colspan="3"><h3>Node pseudo selectors</h3></td></tr>
 <tr><td colspan="3">These selectors enable matching specific leaf nodes, including Comments, TextNodes. When used with `Element#select(String)`, these can be used with structural selectors such as <code>:has()</code> to refine which Elements are matched. To retrieve matching Nodes directly, use `Element#selectNodes(String)`.</td></tr>
 <tr><td>::node</td><td>Matches any node</td><td></td></tr>
 <tr><td>::leafnode</td><td>Matches any leaf-node (this is, a Node which is not an Element)</td><td></td></tr>
 <tr><td>::comment</td><td>Matches a Comment node</td><td></td></tr>
 <tr><td>::text</td><td>Matches a TextNode</td><td></td></tr>
 <tr><td>::data</td><td>Matches a DataNode</td><td></td></tr>
 <tr><td>::node:contains(text)</td><td>Matches a node that has a (normalized, case-insensitive) value containing <i>text</i>.</td><td>::comment:contains(foo bar)</td></tr>
 <tr><td>::node:matches(regex)</td><td>Matches a node that has a value matching the regex.</td><td>::comment:matches(\\d+)</td></tr>
 <tr><td>::node:blank</td><td>Matches a node that has either no value, or a value of only whitespace.</td><td>::comment:not(:blank)</td></tr>
 
 </table>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 * Removed previously deprecated methods. [#2317](https://github.com/jhy/jsoup/pull/2317)
 
 ### Improvements
+* Enhanced the `Selector` to support direct matching against nodes such as comments and text nodes. For example, you can now find an element that follows a specific comment: `::comment:contains(prices) + p` will select `p` elements immediately after a `<!-- prices: -->` comment. Supported types include `::node`, `::leafnode`, `::comment`, `::text`, and `::data`. Node contextual selectors like `::node:contains(text)`, `:matches(regex)`, and `:blank` are also supported. Introduced `Element#selectNodes(String css)` and `Element#selectNodes(String css, Class nodeType)` for direct node selection. [#2324](https://github.com/jhy/jsoup/pull/2324)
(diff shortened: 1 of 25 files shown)


821938089:
Does the Nodes class have a DOM update feature like the Elements class?
If so, please provide aliases for all original array methods.

jhy:
> Does the Nodes class have a DOM update feature like the Elements class? If so, please provide aliases for all original array methods.

@821938089 yup, same as Elements.

atbest:
I think it is better to have an `::cdata` pseudo selector to Match a CDataNode.

jhy:
@atbest makes sense, added

PR #2343, opened 2025-06-22
Deprecate `:matchText` selector

jhy:
The `:matchText` selector has a wonky implementation that requires a specific order to its execution, and fills the DOM up with fake elements. Now that we have the `::textnode` selector and the `Element#selectNodes(CSS, type)` method, we can get rid of `:matchText`.

I've searched around GitHub and other open source projects and have found I think only one actual user of this selector. So the impact to removal is going to be very low.

I included an error log on a user's first use of the selector, so that folks will get a warning of the deprecation. Otherwise as users don't compile against the `MatchText` evaluator, they would not see a warning.

Also added changelog notes and explicit `@Deprecated` annotations to the other methods that are currently deprecated.


anonyein:
Now that the :matchText selector needs to be implemented with Java code, does it become more troublesome?
Can implement a New Reasonable :matchText Using selectNodes("::textnode")

jhy:
@anonyein I don't think I completely follow "needs to be implemented with Java code" -- it always needed code?

The new `public Nodes<Node> selectNodes(Evaluator evaluator)` and `public <T extends Node> Nodes<T> selectNodes(Evaluator evaluator, Class<T> type)` methods are more ergonomic than this deprecated path, in that they return real nodes, and don't mess up the DOM with synthetic `PseudoTextElement` instances.

For any users of `matchText`, moving to that will be a one or perhaps two line change. I don't think that's a particularly high burden. Particularly given the usage is so low.

Regarding your suggestion of re-implementing `:matchText` by backing it with `selectNodes("::textnode")`: this isn't straightforward because the old approach relied on adding synthetic (`PseudoTextElement`) nodes into the DOM. Since we're specifically removing that approach due to its undesirable side-effects, maintaining even a partial compatibility layer wouldn't be very beneficial and would undermine the improvement I'm aiming for.

Issue #2330, opened 2025-05-21
Consider offering a legacy handling of non-xhtml5 compliant self-closing tags?


tballison:
I initially opened https://github.com/jhy/jsoup/issues/2329 for what I thought was a regression. I was wrong, this was a new feature to support the xhtml5 spec.

The more I think about it though, I'm yearning for the old days where messy self-closing tags that were not xhtml5 compliant were closed.

@wiedsche [notes](https://github.com/jhy/jsoup/issues/2329#issuecomment-2898011582) that the new behavior in 1.20.1 can lead to some concerning behavior.

Is there any way to offer a switch for legacy behavior? Aside from xhtml5 compliance, was there anything wrong with the behavior in 1.19.1 and before?

As an open source maintainer, I say from the bottom of my heart, @jhy , that I'm super grateful for all of your work on jsoup. It is a gem!  Thank you.

jhy:
Hi,

I think a good approach for this will be to add the option to supply a tag customizer that is called before vending a tag in `tag#get()` / `tag#valueOf()`. And users can modify as desired.

(Not to be nit-picky but for clarity: jsoup parses [HTML](https://html.spec.whatwg.org/multipage/) (aka HTML5) or XML. It doesn't know about parsing anything called xhtml5. So I've edited the report.)

The previous behavior was incorrect against our goal of following the spec and aligning to current browsers. And in particular could lead to parser differential issues (if jsoup parses differently to how browsers do, that's possibly a path for data / XSS smuggling, depending on how things are configured).

And thanks for the thanks :) It's much appreciated.

wiedsche:
I tested it  with a local snapshot build of 1.21.1 and I can confirm, that this improvement solves my issue with unknown possibly self-closing tags.

Some non-html5-compliant behaving tags, like a self-closing `<strong />`,  will be handled from now on differently within our code, or simply by setting it to be allowed self-closing.

Thanks a lot for the quick support.
I hope to see a release soon.

tballison:
> (Not to be nit-picky but for clarity: jsoup parses [HTML](https://html.spec.whatwg.org/multipage/) (aka HTML5) or XML. It doesn't know about parsing anything called xhtml5. So I've edited the report.)

LOL...I broke my rule of not posting before finishing my coffee. Sorry. Yes, agreed. My bad.

tballison:
Y. That works well. Thank you!

tballison:
[...]

PR #2344, opened 2025-06-22
Update Tag.java

Yuanuo:
Determine if SelfClosing cannot ignore another SeenSelfClose flag

### DIFF
--- a/src/main/java/org/jsoup/parser/Tag.java
+++ b/src/main/java/org/jsoup/parser/Tag.java
@@ -275,7 +275,7 @@ public class Tag implements Cloneable {
      * @return if this tag should be output as self-closing.
      */
     public boolean isSelfClosing() {
-        return (options & SelfClose) != 0 || (options & Void) != 0;
+        return (options & SelfClose) != 0 || (options & SeenSelfClose) != 0 || (options & Void) != 0;
     }
 
     /**


jhy:
Closing as incomplete, failing, and contrary to the design.

See https://github.com/jhy/jsoup/issues/2330#issuecomment-2899689838 for an example allowing self-closing tags by default.

Yuanuo:
Well, let me provide a test code:

```java
package org.jsoup;

import org.jsoup.parser.Parser;
import org.jsoup.parser.Tag;

public class TestNewTag {
        String html = "<p>text 1</p>\n" +
                      "<ol />\n" +
                      "<p>text 2</p>\n" +
                      "<ol />\n" +
                      "<p>text 3</p>";

        Parser parser = Parser.htmlParser();
        parser.tagSet().onNewTag(tag -> {
            if (!tag.isKnownTag())
                tag.set(Tag.SelfClose);
        });

        System.out.println(Jsoup.parse(html, parser).html());
    }
}

```

The result of running this is:

```html
<html>
 <head></head>
 <body>
  <p>text 1</p>
  <ol>
   <p>text 2</p>
   <ol>
    <p>text 3</p>
   </ol>
  </ol>
 </body>
</html>
```
Also, not using the passed parser argument is the same result.

After using my code for this PR, the result of the run is:

```html
<html>
 <head></head>
 <body>
  <p>text 1</p>
  <ol></ol>
  <p>text 2</p>
  <ol></ol>
  <p>text 3</p>
 </body>
</html>
```

I think this should be the right outcome to expect.

PR #2349, opened 2025-06-26
Collapse whitespace between adjacent TextNodes (#2347)

### LINKED ISSUE #2347: Pretty printing consecutive text nodes should collapse consecutive whitespace
Since Jsoup 1.20.1, pretty-printing documents with consecutive text nodes does not fully collapse their whitespace:

```kotlin
val doc = Jsoup.parseBodyFragment("Before <span>Node</span> After")
doc.select("span").remove()
doc.toString()
```

```html
<html>
 <head></head>
 <body>Before  After</body>
</html>
```

(Note the double whitespace between "Before" and "After".)

Re-parsing the document and pretty-printing the result does then fully collapse the whitespace:

```kotlin
val doc = Jsoup.parseBodyFragment("Before <span>Node</span> After")
doc.select("span").remove()
Jsoup.parse(doc.toString()).toString()
```

```html
<html>
 <head></head>
 <body>Before After</body>
</html>
```

In my opinion, the pretty printer should produce the same result for consecutive or merged text nodes. Otherwise, the printer's output is not stable. (I.e. re-parsing and re-printing should not change the result.)

GeorgEchterling:
Fixes #2347

I'm not sure if this can be fixed more elegantly than looping over sequential text nodes. I'd prefer handling multiple text nodes in a single call to `Entities.doEscape`, but that would involve larger code changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # jsoup Changelog
 
+## 1.21.2 (PENDING)
+
+### Improvements
+* When pretty-printing, if there are consecutive text nodes (via DOM manipulation), the non-significant whitespace between them will be collapsed. [#2349](https://github.com/jhy/jsoup/pull/2349). 
+
 ## 1.21.1 (2025-Jun-23)
 
 ### Changes
--- a/src/main/java/org/jsoup/nodes/Printer.java
+++ b/src/main/java/org/jsoup/nodes/Printer.java
@@ -126,20 +126,12 @@ class Printer implements NodeVisitor {
                     options |= Entities.TrimLeading;
             }
 
-            if (next == null || !(next instanceof TextNode) && shouldIndent(next))
-                options |= Entities.TrimTrailing; // don't trim if there is a TextNode sequence
-
-            // do trim trailing whitespace if the next non-empty TextNode has leading whitespace
-            Node subsequentTextNode = next;
-            while (subsequentTextNode instanceof TextNode) {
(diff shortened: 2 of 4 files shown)


GeorgEchterling:
I accidentally auto-formatted a file and introduced a bug during refactoring. These issues should both be fixed now.

I'll take a look at the failing tests. I hadn't run all tests locally because I keep getting compilation errors in the java11 sources:

```
[ERROR] /home/gechterling/dev/open-source/jsoup/src/main/java11/org/jsoup/helper/HttpClientExecutor.java:[3,17] cannot find symbol
  symbol:   class Connection
  location: package org.jsoup
... (and many similar errors)
```

GeorgEchterling:
Nevermind, mvn clean fixed the compilation errors.

GeorgEchterling:
The failing test case was `NodeTest.unwrapNoChildren`. It essentally checked the opposite of my expectation that common whitespace between subsequent text nodes should be collapsed. From the name of the test, I'm assuming that this was not the intended purpose. I've changed the test to expect whitespace to collapse.

jhy:
Thanks, looks good! I simplified the impl to use the existing nextNonBlank method.

> I'd prefer handling multiple text nodes in a single call to Entities.doEscape

We actually used to do that something like that, by peeking into the last character of the Appender, and then not trimming leading. That only worked if the Appender was a StringBuilder. I guess we could track it now on QuiteAppender#append, but it seems likes a lot of overhead. I think this flow works well in the rest of the context of the pretty printer.

PR #2359, opened 2025-07-02
feat: add OutputSettings option to sort attributes during HTML serial…

BrenoJoseCoelho:
…ization

### What does this PR do?
This PR introduces a new configuration option in `Document.OutputSettings`:

public boolean sortAttributes();
public OutputSettings sortAttributes(boolean sort);

By default, attributes are emitted in the order they were added. When sortAttributes is set to true, attributes are sorted by their key before being written to the serialized HTML output.

Why is this useful?
It makes generated HTML output more consistent and predictable.

It helps when diffing HTML output in tests (attribute order can vary otherwise).

It improves reproducibility when serializing elements programmatically.

What changed?
✅ Added sortAttributes field and accessors in Document.OutputSettings.
✅ Updated Attributes.html() to sort attributes when the option is enabled.
✅ Preserved existing behavior when the option is not enabled (default = false).

Example usage
Document doc = Jsoup.parse("<div b='2' a='1'></div>");
doc.outputSettings().sortAttributes(true);
System.out.println(doc.body().html());
// Outputs: <div a="1" b="2"></div>
Tests
(TODO: Add unit test to verify attribute order with sortAttributes = true / false.)

Backward compatibility
✅ No breaking changes
✅ Default behavior remains unchanged unless the new option is explicitly enabled

### DIFF
--- a/src/main/java/org/jsoup/nodes/Attributes.java
+++ b/src/main/java/org/jsoup/nodes/Attributes.java
@@ -12,6 +12,7 @@ import java.util.AbstractSet;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


jhy:
Nonsense untested AI slop. Misses existing output functionality.

I don't accept AI slop PRs into jsoup.

PR #2361, opened 2025-07-03
refactor: add Validate.notNullOrEmpty helper

jparroj:
This PR adds the notNullOrEmpty(String, String) method to the Validate class to consolidate common string validation logic.

✔ Removes duplicated checks for non-null and non-empty strings. ✔ Easier to maintain and reuse.
✔ No functional change, only internal cleanup.


jhy:
jsoup doesn't take or need AI slop PRs.

PR #2364, opened 2025-07-05
Use computeIfAbsent() in TagSet

Isira-Seneviratne:


jhy:
Thanks, but I don't find this to be meaningfully better than what we currently have. I find it harder to follow the flow and reason about the side effects.

PR #2368, opened 2025-07-09
Fix always close resources

backstraw:
Hi,

Previously, resources were generally closed in most cases, but there was an issue where resources were not properly closed when exceptions occurred.
In this PR, try-with-resources was applied to ensure that resources are always safely closed, even if an exception is thrown.

Thanks for reviewing!

### DIFF
--- a/src/main/java/org/jsoup/helper/HttpConnection.java
+++ b/src/main/java/org/jsoup/helper/HttpConnection.java
@@ -1238,8 +1238,14 @@ public class HttpConnection implements Connection {
         }
 
         static void writePost(final HttpConnection.Request req, final OutputStream outputStream) throws IOException {
+            try (OutputStreamWriter osw = new OutputStreamWriter(outputStream, req.postDataCharset());
+                 BufferedWriter w = new BufferedWriter(osw)) {
+                implWritePost(req, w, outputStream);
+            }
+        }
+
+        private static void implWritePost(final HttpConnection.Request req, final BufferedWriter w, final OutputStream outputStream) throws IOException {
             final Collection<Connection.KeyVal> data = req.data();
-            final BufferedWriter w = new BufferedWriter(new OutputStreamWriter(outputStream, req.postDataCharset()));
             final String boundary = req.mimeBoundary;
 
             if (boundary != null) { // a multipart post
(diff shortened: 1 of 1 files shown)


src/main/java/org/jsoup/helper/HttpConnection.java
```diff
@@ -1239,58 +1239,58 @@ else if (needsMultipart(req)) {
 
         static void writePost(final HttpConnection.Request req, final OutputStream outputStream) throws IOException {
             final Collection<Connection.KeyVal> data = req.data();
-            final BufferedWriter w = new BufferedWriter(new OutputStreamWriter(outputStream, req.postDataCharset()));
             final String boundary = req.mimeBoundary;
 
-            if (boundary != null) { // a multipart post
-                for (Connection.KeyVal keyVal : data) {
+            try (BufferedWriter w = new BufferedWriter(new OutputStreamWriter(outputStream, req.postDataCharset()))) {
```
jhy, line 1244: The OutputStreamWriter is not closed -- the BufferedWriter's close does not cascade. I think we should close the OSW also, while we're at it.

jhy, line 1244: The extra nesting here in the try is getting a bit deep, maybe pull out an implWritePost method to outdent this?

jhy (changes requested):

src/main/java/org/jsoup/helper/CookieUtil.java
```diff
@@ -107,14 +107,14 @@ static void storeCookies(HttpConnection.Request req, HttpConnection.Response res
...
-        CharacterReader reader = new CharacterReader(value);
-        String cookieName = reader.consumeTo('=').trim();
-        reader.advance();
-        String cookieVal = reader.consumeTo(';').trim();
-        // ignores path, date, domain, validateTLSCertificates et al. full details will be available in cookiestore if required
-        // name not blank, value not null
-        if (!cookieName.isEmpty())
-            res.cookie(cookieName, cookieVal); // if duplicate names, last set will win
-        reader.close();
+        try (CharacterReader reader = new CharacterReader(value)) {
```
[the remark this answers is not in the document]
backstraw: You're right, this code doesn't throw any exceptions, so try-with-resources may not be necessary here...

src/main/java/org/jsoup/helper/HttpConnection.java
backstraw: Thank you for your feedback! 
I will update the code to explicitly close both the OutputStreamWriter and BufferedWriter, and refactored the main logic into a separate implWritePost method.

jhy:
Thanks, merged

PR #2373, opened 2025-07-21
Performance optimizations for some DOM methods

jhy:
The primary optimization is to invalidate sibling indexes on modification, vs re-calculating them immediately. When called in a loop of `element.child(0).remove()`, the previous implementation was effectively O(n²), and is now just O(n).

Also includes optimizations for `Document.createElement()`, setting attributes, and when parsing a body fragment with many direct children.

### DIFF
--- a/src/main/java/org/jsoup/internal/StringUtil.java
+++ b/src/main/java/org/jsoup/internal/StringUtil.java
@@ -148,8 +148,8 @@ public final class StringUtil {
      * @param string string to test
      * @return if string is blank
      */
-    public static boolean isBlank(final String string) {
-        if (string == null || string.length() == 0)
+    public static boolean isBlank(@Nullable String string) {
+        if (string == null || string.isEmpty())
             return true;
 
         int l = string.length();
--- a/src/main/java/org/jsoup/nodes/Attributes.java
+++ b/src/main/java/org/jsoup/nodes/Attributes.java
@@ -159,7 +159,7 @@ public class Attributes implements Iterable<Attribute>, Cloneable {
         if (i != NotFound)
             vals[i] = value;
         else
-            add(key, value);
+            addObject(key, value);
         return this;
     }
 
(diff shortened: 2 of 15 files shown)


jhy:
@carlosame FYI, this change corrects the busted performance of jsoup in your [DOM modification benchmark](https://css4j.github.io/dom-mark.html#dom-change) ([code](https://github.com/css4j/benchmark/blob/2d82dd1d56ada76f93cfcc2574854a7865f3cbbc/src/io/sf/carte/mark/dom/DOMChangeMark.java)).

New results:

```text
Benchmark                       Mode  Cnt     Score    Error  Units
DOMChangeMark.markChangeDOM    thrpt    3  1539.337 ± 63.919  ops/s
DOMChangeMark.markChangeDOM4J  thrpt    3   363.892 ± 55.999  ops/s
DOMChangeMark.markChangeJdk    thrpt    3  2736.017 ± 83.992  ops/s
DOMChangeMark.markChangeJsoup  thrpt    3   366.781 ± 10.708  ops/s
```

Where previously jsoup was a pathetic 6 ops/s on this machine. As mentioned in the commit, the flow of the benchmark's removal was running as O(n²); now it's O(n).

From a quick look at the JDK's DOM impl, the primary difference that gives the extra speed for this specific benchmark is that they are using a linked-list vs an array, and so the construction/removal is just a simple field change. And as it's XML they're not needing to lowercase normalize the tag names or attributes etc. There are pros and cons to using a linked list in a DOM; for jsoup's general use case I've found the list is better, but I'll noodle on that some more.

carlosame:
> From a quick look at the JDK's DOM impl, the primary difference that gives the extra speed for this specific benchmark is that they are using a linked-list vs an array, and so the construction/removal is just a simple field change. And as it's XML they're not needing to lowercase normalize the tag names or attributes etc. There are pros and cons to using a linked list in a DOM; for jsoup's general use case I've found the list is better, but I'll noodle on that some more.

Thanks for the heads-up, I'll update the benchmarks once 1.21.2 is out.

Css4j's native DOM uses a linked list but has all the HTML overhead, it scores 1539 in the benchmark that you ran. The JDK DOM used to be about 55% faster than css4j DOM but has improved in recent JDKs. The JDK is generally faster because it does some weird internal optimizations; those sometimes trigger errors (_e.g._ `ArrayIndexOutOfBoundsException`) during thread-heavy benchmarks.

PR #2403, opened 2025-09-22
Java 25 support

offa:
Adds Java 25 CI builds.

There's a test error though (Java 25 on Linux only):

```
Error:  org.jsoup.integration.ProxyTest.canAuthenticateToProxy(String)[2] -- Time elapsed: 0.024 s <<< FAILURE!
org.opentest4j.AssertionFailedError: Expected 407 Proxy Authentication Required, got: Cannot invoke "jdk.internal.net.http.ExchangeImpl.cancel(java.io.IOException)" because "exch.exchImpl" is null
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
	at org.junit.jupiter.api.Assertions.fail(Assertions.java:138)
	at org.jsoup.integration.ProxyTest.assertAuthRequiredException(ProxyTest.java:149)
	at org.jsoup.integration.ProxyTest.canAuthenticateToProxy(ProxyTest.java:129)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:186)
[…]
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### Improvements
 * Added an instance method `Parser#unescape(String, boolean)` that unescapes HTML entities using the parser’s configuration (e.g. to support error tracking), complementing the existing static utility `Parser.unescapeEntities(String, boolean)`. [#2396](https://github.com/jhy/jsoup/pull/2396)
+* Build: added CI coverage for JDK 25 [#2403](https://github.com/jhy/jsoup/pull/2403)
 
 ### Bug Fixes
 * Previously cached child Elements of an Element were not correctly invalidated in `Node#replaceWith(Node)`, which could lead to incorrect results when subsequently calling `Element#children()`. [#2391](https://github.com/jhy/jsoup/issues/2391)


offa:
All three are using Azul Java 25, but only one fails :thinking:.

jhy:
Deeper stack is:

```text
Caused by: java.lang.NullPointerException: Cannot invoke "jdk.internal.net.http.ExchangeImpl.cancel(java.io.IOException)" because "exch.exchImpl" is null
	at java.net.http/jdk.internal.net.http.MultiExchange.lambda$responseAsyncImpl$0(MultiExchange.java:446)
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1171)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:531)
	at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2206)
	at java.net.http/jdk.internal.net.http.ResponseSubscribers.lambda$getBodyAsync$1(ResponseSubscribers.java:1090)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:884)
	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:862)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:531)
	at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2206)
```

Looking at jdk.internal.net.http.Exchange, I see that exchangeImpl is volatile and the code has various mentions of possible races when cancelling the request, and a lot of null guards, but not around where this fires. So far as I can see, jsoup is calling correctly; and it just happens that on this platform something is racing slower/faster than others and not producing the correct error.

I will try just having a specific catch for that string. Really the tests are just making sure that the proxy is actually checking authentication; as long as we get an exception on the incorrect password, and a pass on the correct, the test logic should be satisfied.

jhy:
Thanks @offa, merged.

I think we can run with the four versions for a few weeks and then probably will remove 21.

offa:
Thanks for fixing the exception!

jhy:
(I've removed 21 in eeae0ff3bf4c34565d422ede2b4aee8e6ab3656c)

PR #2420, opened 2025-10-17
Use isEmpty to check whether a String is empty or not

mridul-hossain:


jhy:
What, not even a "whoops"?

Issue #2416, opened 2025-10-13
Align handling of too deep documents with browsers


nicolo-ribaudo:
Consider this test case:

<details>

```html
<html>

<head>
  <title>jsoup bug</title>
  <meta content="text/html; charset=UTF-8" http-equiv="Content-Type">
</head>

<body>
  <div style="display: none">
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
    <span>
[...]

Issue #955, opened 2017-10-18
Quadratic behaviour on deeply nested pages


usamec:
On pages with very deep sequence of elements (like this one sv.stargate.wikia.com/wiki/M2J), Jsoup gets very slow and spends too much time in this function:
https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/parser/HtmlTreeBuilder.java#L462

Is there any way to remove this quadratic behaviour? Either by using better data structures or by having option to limit stack size (and throw exception when it is too deep).

jhy:
Interesting, looks like Chrome has a bit of a perf issue on that page too. Performance monitor shows it continually parsing the page.

Maybe we can implement an improvement here by just running a counter on the tags that in scope. Or put them in a hash or something - vs walking through the scope stack each time and hunting for them.

jhy:
Thanks. I fixed this by limiting the search scope depth to 100. I sampled a bunch of sites in the wild from Alexa 1MM (though could be more rigorous) and found the max was 26. So this depth should be safe. Parse completes in ~ 200 millis now.

Also tuned the string array searches to use binary searches.

PR #2421, opened 2025-10-21
Align maximum HTML depth handling with browsers

nicolo-ribaudo:
This patch aligns the way jsoup handles too deep DOM trees with what browsers do. Note that all browsers have different behaviors, so I picked what seemed to be most reasonable/implementable.

|                                                                                                                                   | jsoup | Chrome | Firefox | Safari |
|-----------------------------------------------------------------------------------------------------------------------------------|-------|--------|---------|--------|
| Maximum depth<br>https://github.com/whatwg/html/issues/3732#issuecomment-3400961504                                                                                                                    | 512   | 513    | 513     | 512    |
| Implicit elements inside `<table>`<br>allow going past the max depth<br>https://github.com/whatwg/html/issues/3732#issuecomment-3422993523                                                                | No    | Yes    | No      | Yes    |
| A closing tag after an opening tag<br>that is auto-closed due to the max<br>depth will close the previous matching<br>element on the stack<br>https://github.com/whatwg/html/issues/3732#issuecomment-3426984910 | Yes   | No     | No      | Yes    |

Ref #2416 (let's keep that issue open to track HTML spec changes related to this?) 

I actually think that the last line would be more intuitive as "Yes", but that would cause the same perf issues that were fixed by MaxScopeSearchDepth (since we'd need an unlimited stack size).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,9 @@ updates:
         update-types: ["version-update:semver-major"]
       - dependency-name: "org.eclipse.jetty:jetty-servlet"
         update-types: ["version-update:semver-major"]
+      # Et tu, junit? Keep us on 5, as 6 has min JDK17 - https://docs.junit.org/6.0.0-RC3/release-notes/#release-notes-6.0.0-M1
+      - dependency-name: "org.junit.jupiter:junit-jupiter"
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: github-actions
     directory: /
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -20,11 +20,11 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
       - name: CodeQL Initialization
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@v4
         with:
           languages: java
           queries: +security-and-quality
       - name: Autobuild
-        uses: github/codeql-action/autobuild@v3
+        uses: github/codeql-action/autobuild@v4
       - name: CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@v4
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,8 @@
 * Null characters in the HTML body were not consistently removed; and in foreign content were not correctly replaced. [#2395](https://github.com/jhy/jsoup/issues/2395)
 * An IndexOutOfBoundsException could be thrown when parsing a body fragment with crafted input. Now logged as a parse error. [#2397](https://github.com/jhy/jsoup/issues/2397), [#2406](https://github.com/jhy/jsoup/issues/2406)
 * When using StructuralEvaluators (e.g., a `parent child` selector) across many retained threads, their memoized results could also be retained, increasing memory use. These results are now cleared immediately after use, reducing overall memory consumption. [#2411](https://github.com/jhy/jsoup/issues/2411)
+* Cloning a `Parser` now preserves any custom `TagSet` applied to the parser. [#2422](https://github.com/jhy/jsoup/issues/2422), [#2423](https://github.com/jhy/jsoup/pull/2423)
+* Custom tags marked as `Tag.Void` now parse and serialize like the built-in void elements: they no longer consume following content, and the XML serializer emits the expected self-closing form. [#2425](https://github.com/jhy/jsoup/issues/2425)
 
 ### Internal Changes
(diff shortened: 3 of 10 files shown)


nicolo-ribaudo:
About the CI failure: should I keep `MaxScopeSearchDepth`, even if it's now unused?

jhy:
Thanks for this. First pass responses:

> About the CI failure: should I keep MaxScopeSearchDepth, even if it's now unused?

Yes for now we should keep it in and mark it `@Deprecated` so that we can remove. I doubt it's used by anyone but that gives us a path. 

> I actually think that the last line would be more intuitive as "Yes", but that would cause the same perf issues that were fixed by MaxScopeSearchDepth (since we'd need an unlimited stack size).

Could you clarify that -- you do have it marked as "yes" ? Which do you think is best, and is there an impact if that is different from Chrome as predominant browser? We are trying to align to that in lieu of a followed spec, right? And I don't follow the point on requiring an unlimited stack to implement that -- aren't we just walking up the limited set?

nicolo-ribaudo:
> Could you clarify that -- you do have it marked as "yes" ? Which do you think is best, and is there an impact if that is different from Chrome as predominant browser? We are trying to align to that in lieu of a followed spec, right? And I don't follow the point on requiring an unlimited stack to implement that -- aren't we just walking up the limited set?

Chrome/Firefox do not actually have a stack size limit, and that's how they are able to match opening/closing element _at any depth_. The have a limit on the _dom tree depth_. The can to see 10'000 alternating open `<div>`s and `<span>`s, 9'999 closing ones, and be able to tell that only the outermost one was left open.

I personally think that the Chrome/Firefox behavior is better than Safari, but Safari has 15%-20% market share so it's probably reasonable to expect that pages work there, and thus their HTML is something that behaves the same across Chrome and Safari.

Having an unlimited stack size in jsoup was not desirable for perfomance reasons, because it quadratically iterates through that stack (https://github.com/jhy/jsoup/issues/955).

If you want, I can try to have an unlimited stack size while still avoiding the quadratic behavior, however it comes at an extra memory cost. We'd need to have, next to the stack, a map of (element name)->(stack of indexes that that element has in the stack), so that we could query "where does this element appear in the stack" in O(1) time.

nicolo-ribaudo:
> If you want, I can try to have an unlimited stack size while still avoiding the quadratic behavior, however it comes at an extra memory cost. We'd need to have, next to the stack, a map of (element name)->(stack of indexes that that element has in the stack), so that we could query "where does this element appear in the stack" in O(1) time.

EDIT: This is much more complex than just having that map, because of all the searches we need to do in the stack "check if there is any element of type x/y/z but stop when you find a/b/c".

nicolo-ribaudo:
You can see for example Firefox's implementation, which is very similar to Chrome's but it's written in Java so it's easy to compare:

- they have an upper limit when getting the "last" element of the stack to know to which parent insert the new element (https://github.com/validator/htmlparser/blob/b19d4088f1d138715a981e960f0720678e7946f7/src/nu/validator/htmlparser/impl/TreeBuilder.java#L6349)
- they do not have any limit when iterating the stack (see https://github.com/validator/htmlparser/blob/b19d4088f1d138715a981e960f0720678e7946f7/src/nu/validator/htmlparser/impl/TreeBuilder.java#L4168 and similar methods)
- they do not have any limit when pushing elements to the stack (https://github.com/validator/htmlparser/blob/b19d4088f1d138715a981e960f0720678e7946f7/src/nu/validator/htmlparser/impl/TreeBuilder.java#L4556)

jhy:
I'm confused as to those [HTML Validator](https://about.validator.nu/htmlparser/) source links. That's not the Firefox parser and I don't see that we can use it as a reference for anything.

Firefox code is in this [repo](https://searchfox.org/firefox-main/source/parser/html), per this [doc](https://firefox-source-docs.mozilla.org/contributing/directory_structure.html).

nicolo-ribaudo:
That C++ is generated by transpiling the Java code: https://searchfox.org/firefox-main/source/parser/html/java/README.txt

jhy:
But the TreeBuilder hasn't changed in seven years? How can that be?

jhy:
I have added some changes so that
- the XML parser also can apply a limit. By default it's unlimited / max integer
- when we pop to limit, we also clean up the other tree builder state (form, head elements, template, format els)

nicolo-ribaudo:
> But the TreeBuilder hasn't changed in seven years? How can that be?

For some reason I linked to the commit where the changes to the depth handling were first introduced, but the last commit is from two months ago.

jhy:
Ah sorry, I completely missed that I wasn't viewing on head there. Thanks again for the report detail and the PR.

PR #2440, opened 2025-12-03
Refactor: improve parameter names in boolean validation helpers

brenot0:
Renamed parameters in isTrue/isFalse methods from val/msg to condition/message to improve readability, without changing behavior.


jhy:

PR #2441, opened 2025-12-03
Refactor: extract helper method for Element#child

brenot0:
Extracted the loop that locates a child element by filtered index into a private helper method (childByFilteredIndex), improving readability while keeping the behavior unchanged.


jhy:

PR #2442, opened 2025-12-03
Docs: refine Javadoc for Element.select(String)

brenot0:
Refined the Javadoc of Element#select(String cssQuery) to better describe behavior, parameters and examples of usage, without changing implementation.


jhy:

PR #2453, opened 2025-12-21
Isolate TagSet copies to prevent shared mutation

jhy:
TagSet copies no longer pull through from the mutable template; they copy existing tags and only use the template’s fallback (the HTML defaults) for lazy lookups. This keeps parser/session tagsets independent and prevents cross-thread mutation that can lead to ConcurrentModificationException.

This keeps the use of the HTML default tagset sparse in memory.


jhy:
Was seeing occasional CI breaks in SessionIT (e.g. https://github.com/jhy/jsoup/actions/runs/20250357192/job/58140769255), this aims to fix those.

PR #2456, opened 2025-12-29
Validate change string check function

devch37:
Hello 😄 

Replaced string.length() == 0 with string.isEmpty() for improved readability and clearer intent.

Thanks for your open source !


jhy:
Thanks, but I think it's fine as is.

PR #2494, opened 2026-04-22
Add Elements.before(Node) with tests

### LINKED ISSUE #953: Add method Elements.before(Node node)

echu921:
This PR adds support for Elements.before(Node node), extending the existing API to allow inserting a Node before each element in an Elements collection.

Currently, jsoup supports:
- Element.before(Node node) for single elements
- Elements.before(String html) for collections

However, there is no equivalent method for inserting a Node across an Elements collection. This requires users to convert nodes to HTML strings (e.g., via outerHtml()), which is less intuitive and less efficient.

This implementation applies the insertion to each element in the collection and clones the provided node to ensure correct DOM behavior when inserting before multiple elements.

Tests were added to verify:
- insertion before multiple elements
- correct cloning behavior
- empty element collections
- null input handling

Closes #953

### DIFF
--- a/src/main/java/org/jsoup/select/Elements.java
+++ b/src/main/java/org/jsoup/select/Elements.java
@@ -342,6 +342,22 @@ public class Elements extends Nodes<Element> {
     }
 
     /**
+     Insert the supplied node before each matched element's outer HTML.
+
+     @param node node to insert before each element
+     @return this, for chaining
+     @see Element#before(Node)
+     */
+    public Elements before(Node node) {
+        Validate.notNull(node);
+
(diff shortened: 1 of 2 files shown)


jhy:
Please read the contributing guidelines. The jsoup project does not accept LLM contributions.

PR #2498, opened 2026-05-04
Refactor source range storage

jhy:
I've refactored how jsoup retains source ranges, so that enabling parser position tracking has a much smaller retained-memory cost.

The issue was mostly in how the data was stored: source ranges were being kept in node and attribute user data. That meant that the range track data took the same relatively heavy path as normal user data: leaf nodes could be expanded into `Attributes` just to hold range data, and attribute ranges would be backed by adding user data hash maps. On documents with a lot of attributes, that adds up quickly.

This PR keeps the public range-reading API intact, but changes the internal storage model:

- `Range` now stores source offsets, and derives `Position` line / column values from a parse-level line map.
- `LineMap` records newline offsets once per parse, instead of retaining line and column numbers in every range position.
- `Range.Spans` is the compact internal record for node start ranges, element end ranges, and attribute ranges.
- `LeafNode` keeps its compact value storage, and wraps the value in a TrackedValue when it also needs retained source range data, vs setting up extra Attributes
- Attribute ranges are stored by directly in internal attribute slots, avoiding per-attribute user data maps.
- The old parser-internal setup methods are retained for source compatibility, but deprecated and guarded so ranges from different parses cannot be mixed into the same stored span.

This also simplifies the parser data path and lowers GC. During parsing we now pass primitive source offsets into the retained span storage, instead of constructing `Range` and `Position` objects as intermediate storage. `Position` objects are created lazily when user code asks for `range.start()` or `range.end()`, so retained documents do not carry that object graph and normal parse flow does less short-lived allocation.

Retained DOM footprint, on JDK 21.0.10:

| Fixture | Input | # Nodes | # Attrs | Retained DOM | 1.22.2 tracked | 1.23.1 tracked | Saving |
|---|---:|---:|---:|---:|---:|---:|---:|
| tiny | 274 B | 23 | 5 | 30.2 KB | 40.0 KB | 32.1 KB | 19.7% |
| small | 3.3 KB | 112 | 31 | 41.7 KB | 88.8 KB | 50.7 KB | 42.9% |
| medium | 8.6 KB | 397 | 121 | 73.4 KB | 244.4 KB | 103.9 KB | 57.5% |
| large | 496.3 KB | 18,497 | 6,152 | 1.9 MB | 9.6 MB | 3.3 MB | 65.4% |
| x-large | 2.3 MB | 26,364 | 23,244 | 5.2 MB | 19.4 MB | 7.3 MB | 62.5% |

The saving column compares tracked DOM retained size in 1.22.2 against tracked DOM retained size with this change. Retained DOM is shown as the no-source-tracking baseline.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## 1.23.1 (pending)
 
+### Improvements
+* Reduced retained memory when parsing with source position tracking enabled (`Parser#setTrackPosition(true)`). Source ranges are now stored in compact parser-owned span records instead of node and attribute user data, and `Position` objects are created lazily when source ranges are read. This cuts tracked DOM retained size by about 50-60% on representative benchmark documents, while keeping `Node#sourceRange()`, `Element#endSourceRange()`, and `Attribute#sourceRange()` behavior intact. [#2498](https://github.com/jhy/jsoup/pull/2498)
+
 ### Build Changes
 * Cleaned up the Maven build for the multi-release JAR so Java 8 and Java 11+ sources compile as separate source sets. This avoids spurious Java 8 compiler warnings from newer-language overlay sources, keeps long-running parser checks behind an explicit profile, and preserves the same published artifacts and runtime behavior.
 
--- a/src/main/java/org/jsoup/internal/LineMap.java
+++ b/src/main/java/org/jsoup/internal/LineMap.java
@@ -1,0 +1,79 @@
+package org.jsoup.internal;
+
+import java.util.Arrays;
+
+/**
+ Maps source offsets to line and column coordinates. Jsoup internal; API subject to change.
+ */
+public final class LineMap {
+    private static final int InitialLineCapacity = 16;
+    private static final int CapacityGrowthFactor = 2;
+    private static final int[] Empty = new int[0];
+    private int[] lineStarts = Empty;
+    private int size;
+
+    /**
+     Creates a shared line map for tracked source ranges.
+     */
+    public LineMap() {}
+
+    /**
+     Records the source offset immediately after a newline.
+     */
+    public void addLineStart(int pos) {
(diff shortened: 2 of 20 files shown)


jhy:
> CodeQL found more than 20 potential problems in the proposed changes. Check the [Files changed](/jhy/jsoup/pull/2498/files) tab for more details.

Which were all just warnings about our maintained tests against the deprecated internal APIs.

jhy:
Here's two benchmarks to compare parse throughput (parse operations per second) and GC rate (bytes per operation). 'untracked' is without source tracking enabled (and was not impacted by this PR). This is a meaningful improvement across each metric if you're using the source range tracking.

**Throughput**

| Fixture | untracked ops/s | 1.22.2 tracked ops/s | 1.23.1 tracked ops/s | Tracked speedup |
|---|---:|---:|---:|---:|
| tiny | 444,978 | 226,387 | 343,712 | +51.8% |
| small | 55,137 | 31,252 | 41,676 | +33.4% |
| medium | 24,007 | 10,448 | 16,215 | +55.2% |
| large | 507.0 | 202.1 | 330.2 | +63.4% |
| x-large | 145.2 | 67.6 | 95.5 | +41.2% |

**GC**

| Fixture | untracked B/op | 1.22.2 tracked B/op | 1.23.1 tracked B/op | Allocation reduction |
|---|---:|---:|---:|---:|
| tiny | 8,056 | 19,568 | 10,200 | 47.9% |
| small | 18,504 | 80,384 | 28,328 | 64.8% |
| medium | 46,832 | 277,280 | 81,400 | 70.6% |
| large | 2,149,647 | 12,888,809 | 3,838,531 | 70.2% |
| x-large | 10,580,426 | 31,376,587 | 13,750,944 | 56.2% |

(Single threaded performance; throughput for relative comparison)

PR #2551, opened 2026-08-02
Make progressive StreamParser selections emission-aware

jhy:
`StreamParser` progressively builds a DOM and emits elements as they become ready, allowing large inputs to be processed without first parsing the whole document.

`selectFirst(...)` searches the document built so far. If it does not find a match there, it continues by running the parser until a matching element reaches its normal stream-emission point.

Given:

```html
<title>One</title><p id=hit>Full</p><p>Next</p>
```

calling `selectFirst("#hit")` directly already returned the complete paragraph:

```html
<p id="hit">Full</p>
```

Consecutive selections, however, could return an incomplete result:

```java
Element title = parser.selectFirst("title");
Element hit = parser.selectFirst("#hit");
```

Emitting `title` advances the parser far enough to create the following paragraph as lookahead, but not to parse its contents. The second selection found that paragraph in the progressive document and treated its presence as readiness, returning:

```html
<p id="hit"></p>
```

`selectFirst(...)` now advances the provisional match to that emission point and returns:

```html
<p id="hit">Full</p>
```

Consecutive selections may consume slightly more input before returning, but callers receive the match with its parsed contents without switching to `selectNext(...)` or completing the document. Later input remains unparsed until needed.

---

As background: HTML parsing does not construct the DOM in simple start-tag order.

The tree builder creates and closes implicit elements, foster-parents malformed table content, reconstructs formatting elements, and can change node parentage as input arrives. StreamParser’s iterator accommodates this by emitting children before their parents and, where applicable, retaining an element until its next sibling is known.

Consequently, `selectFirst(...)` cannot simply return the first match currently present in the progressive document. It also cannot just return the first iterator match because iterator emission order is not always document order.

Selectors such as `:empty` add another wrinkle because an element can match provisionally and stop matching as its contents arrive.

---

Readiness is now derived from TreeBuilder’s explicit open-element and completion state together with StreamParser’s iterator state.

`selectFirst(...)` now follows three stages:

1. Search the progressive document and advance an existing provisional match if necessary.
2. Use the iterator to make linear progress until a ready match is emitted.
3. Reconcile the emitted match with the progressive document, waiting on the current document-first pending candidate where necessary.

Pathological malformed input can cause HTML recovery algorithms to reparent an open descendant beneath an element that previously appeared ready. Proving that a subtree can never change would require parsing the entire remaining input, defeating StreamParser’s purpose.

This change therefore retains StreamParser’s normal emission point as its practical readiness boundary. It improves progressive selection without turning `selectFirst(...)` into an implicit full-document parse.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## 1.23.2 (pending)
 
+### Improvements
+* Improved consecutive `StreamParser.selectFirst(...)` calls during progressive parsing, so later matches are returned with their parsed contents when earlier selections have left them as parser lookahead. E.g., given `<title>One</title><p id=hit>Full</p><p>Next</p>`, selecting `title` and then `#hit` now advances the partial lookahead and returns `<p id="hit">Full</p>`, rather than returning an empty `<p id="hit"></p>` before its content is parsed. The updated readiness tracking follows StreamParser's normal emission order across implicit HTML structure and parser recovery.
+
 ### Bug Fixes
 * Fixed the JDK `HttpClient` implementation to accept responses without a `Content-Type` header, matching the `HttpURLConnection` implementation. [#2549](https://github.com/jhy/jsoup/pull/2549)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ## 1.23.2 (pending)
 
 ### Improvements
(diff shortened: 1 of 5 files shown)


nazimuddin5977-arch (commented):
@nazimuddin5977-arch

PR #2556, opened 2026-08-09
Optimize XML namespace scope tracking

jhy:
Previously, XmlTreeBuilder copied the complete in-scope namespace map for every start tag. As nested declarations grew that map, parsing retained progressively larger copies and spent increasing time recreating them.

This keeps one current bindings map instead, recording only the bindings changed by each element and restoring their parent values when it closes. This makes namespace scope tracking linear in time and retained state while preserving namespace resolution and fragment inheritance.


jhy:
Thanks to @MegaManSec for reporting this!

jhy:
I would note that CVE-2026-75140 was lodged by a third-party CNA, not jsoup, and I disagree with the report. jsoup is a library, not a framework, and as such implementors need to manage thread lifecycles themselves. In my testing when working on https://github.com/jhy/jsoup/pull/2556 with an adversarial payload, a simple try / catch on parser threads (which any standard concurrency framework will implement) was suitable to catch allocation failures and to allow other threads to make consistent progress.

At any rate I do encourage folks to upgrade as there are many good improvements in the release.

PR #2567, opened 2026-08-13
Add Elements.before(Node) and after/prepend/append(Node) #953

### LINKED ISSUE #953: Add method Elements.before(Node node)

AzazelSensei:
#953 asked for Elements.before(Node). You said to do the same for after, prepend, and append so they match the string methods.

Each call clones the node. That way inserting into a list does not move one instance between targets, and I am not serializing through HTML.

Tests are in ElementsTest next to the existing string cases.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,11 +3,11 @@
 ## 1.23.2 (pending)
 
 ### Improvements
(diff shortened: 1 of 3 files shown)


jhy:
Thanks, merged

PR #2584, opened 2026-08-24
Java 11 HTTP client reuse covers ordinary connections

### LINKED ISSUE #2516: Review HttpClient's thread use under heavy request load
#2515 reports that many repeated requests will spawn many threads, probably redundant. They do get reaped on GC pressure, but there is likely to be better ways to reuse them via a keep-alive pool or similar.

@rs1gg thanks for raising this!

jhy:
Requests via `Jsoup.newSession()` already reused their HttpClient. This extends that reuse to independent `Jsoup.connect(...)` calls, so repeated requests do not create a new transport client and its threads each time.

Custom authentication and SSL contexts remain isolated, and proxy selection remains per request.

Fixes: #2516


jhy:
Both runs used JDK 21.0.12, 200 requests, and concurrency 32 where applicable.

| Workload | Server connections | HttpClient threads |
|---|---:|---:|
| Independent sequential requests | 200 → 1 | 379 → 4 |
| Session sequential requests | 1 → 1 | 4 → 4 |
| Independent concurrent requests | 200 → 32 | 357 → 40 |
| Session concurrent requests | 63 → 32 | 126 → 44 |

So this really improves both thread / GC, and ability for connection keep-alive, when using independent requests vs session requests.

PR #2610, opened 2026-09-17
Improve :has() performance and correctness

jhy:
We now search only the children, following siblings, or sibling subtrees that can match each relative selector.

This avoids unnecessary traversal, and fixes sibling alternatives and sibling-descendant selectors that previously returned no match.


jhy:
I ran the sibling `:has()` benchmarks against the large [jsoup API index](https://jsoup.org/apidocs/index-all) page (508KB). The document contains 6K+ elements, with sibling groups up to 336 elements wide. Both selectors match 1,271 `dt` elements.

| Selector | Before | After | Speedup | Allocation |
|---|---:|---:|---:|---:|
| `dt:has(+ dd)` | 398 ops/s | 4,776 ops/s | 12.0x | -98.0% |
| `dt:has(~ dd)` | 19 ops/s | 4,758 ops/s | 251.1x | -99.2% |

Previously, sibling `:has()` queries started at the subject's first sibling then used structural evaluators to establish the relationship for each candidate.

The parsed selector group now retains its top-level combinators, so `QueryParser` can choose an optimized matching traversal directly.

`:has(+ dd)` checks only the next element sibling, while `:has(~ dd)` starts with the following siblings.

Since the traversal already establishes the sibling relationship, the evaluator only needs to test whether each candidate is a `dd`.

